PR #11, opened 2026-01-06
Added a unified way to get input events called get_input_event() 

DallinTG:
it will return a struct that contains the key/mousebutton/gamepad input data this should make it easier to add a system to handle custom keybinds and support for keyboard/gamepad

Adds these data structures ::

Input_Event ::union{
	Keyboard_Key,
	Mouse_Button,
	Mouse_Actions,
	Gamepad_Axis,
	Gamepad_Button,
}

//this is the data a Input_Event could return
Input_Event_Data::struct{
	event:Input_Event,// the Input_Event that created this struct (key/button)
	went_down:bool,
	went_up:bool,
	is_held:bool,
	value:f32,//used for Gamepad_Axis,mouse_wheel_delta
	vector:[2]f32,//used for mouse_position, mouse_delta
}

//this can be used to get related Input_Event_Data
Mouse_Actions :: enum{
	mouse_position,
	mouse_delta,
	mouse_wheel_delta,
}

get_input_event() // this is the proc 

Input_Event_Data is what gets returned if the input event does not use a field it is left at zero

PSS: sory about the massive dif i was trying out zed when i did this and i did not realise it removed trailing white spaces on save

### DIFF
--- a/karl2d.doc.odin
+++ b/karl2d.doc.odin
@@ -188,7 +188,6 @@ get_gamepad_axis :: proc(gamepad: Gamepad_Index, axis: Gamepad_Axis) -> f32
 // not vibrate with the same speed.
 set_gamepad_vibration :: proc(gamepad: Gamepad_Index, left: f32, right: f32)
 
-
 // Returns the stae of a specified input_event(key/mouse_button/game_pad)
 // Gamepad_Index only maters if you are targeting a gamepad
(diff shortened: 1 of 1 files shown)


karl-zylinski:
Thanks for the PR.

I'm closing this due to:
- I'm afraid I don't like how the concept is designed.
- There are too many random whitespace changes.
- The code style is inconsistent.

PR #21, opened 2026-01-07
Native Wayland support

jqcorreia:
This PR implements basic Wayland support.

Depends on:
- libwayland (bindings vendored in linux/wayland)
- EGL (included Odin bindings)

The flow is highly dependent on the init() function for the platform code whereas the _gl_get_context() only calls egl.MakeCurrent().

### DIFF
--- a/window_cocoa.odin
+++ b/window_cocoa.odin
@@ -14,6 +14,7 @@ WINDOW_INTERFACE_COCOA :: Window_Interface {
 	shutdown = cocoa_shutdown,
 	window_handle = cocoa_window_handle,
 	process_events = cocoa_process_events,
+	after_frame_present = cocoa_after_frame_present,
 	get_events = cocoa_get_events,
 	get_width = cocoa_get_width,
 	get_height = cocoa_get_height,
@@ -244,6 +245,10 @@ cocoa_process_events :: proc() {
 			s.app->sendEvent(event)
 		}
 	}
+}
+
+cocoa_after_frame_present :: proc () {
+	
 }
 
 cocoa_get_events :: proc() -> []Window_Event {


jqcorreia:
@karl-zylinski

render_backend_gl_linux.odin
```diff
@@ -5,69 +5,119 @@ package karl2d
...
 import "log"
+import "vendor:egl"
+import "core:fmt"
 
 _ :: log
 
-GL_Context :: struct {
+GL_Context :: union {
+    GL_Context_GLX,
+    GL_Context_EGL
```
karl-zylinski: Split EGL and GLX context creation up if needed

window_interface_chooser.odin
```diff
@@ -5,7 +5,8 @@ when ODIN_OS == .Windows {
 } else when ODIN_OS == .JS {
 	DEFAULT_WINDOW_INTERFACE :: WINDOW_INTERFACE_JS
 } else when ODIN_OS == .Linux {
-	DEFAULT_WINDOW_INTERFACE :: WINDOW_INTERFACE_X11
+	// DEFAULT_WINDOW_INTERFACE :: WINDOW_INTERFACE_X11
+	DEFAULT_WINDOW_INTERFACE :: WINDOW_INTERFACE_WAYLAND
```
karl-zylinski: I can try doing the "dynamic chooser" code. If we merge this PR, then we can do it with X11 as default and I can write the chooser code + a linux "platform abstraction"

window_wayland.odin
```diff
@@ -0,0 +1,509 @@
...
+        egl.NONE,
+    }
+    egl_display := egl.GetDisplay(egl.NativeDisplayType(display))
+    if egl_display == egl.NO_DISPLAY {
+        panic("Failed to create EGL display")
+    }
+    if !egl.Initialize(egl_display, &major, &minor) {
+        panic("Can't initialise egl display")
+    }
+    if !egl.ChooseConfig(egl_display, raw_data(config_attribs), &egl_config, 1, &n) {
```
karl-zylinski: It seems strange to me that there is any egl code at all in the wayland window file. Isn't it possible to just get a window handle and move all the egl stuff to the render_back_gl_linux.odin file?

window_wayland.odin
```diff
@@ -0,0 +1,509 @@
...
+
+
+global :: proc "c" (
+	data: rawptr,
+	registry: ^wl.wl_registry,
+	name: c.uint32_t,
+	interface: cstring,
+	version: c.uint32_t,
+) {
+	context = runtime.default_context()
```
karl-zylinski: Don't use `runtime.default_context`. It doesn't have any of the loggers etc set up. It's a bit iffy what to do instead when it comes to callbacks. You could save away the context in the Wayland_State when you add `registry_listener ` and restore it here.

karl-zylinski (changes requested):
Thanks for the PR, I left some comments about the major things I saw.

window_wayland.odin
jqcorreia: I don't like it either, but the Wayland gods are picky about this. I will try to move the EGL instantiation code to render_back_gl_linux.odin but I then either use state globally on the callbacks or I need to have egl objects present in the WindowHandle.

window_wayland.odin
jqcorreia: I use this as a way to have quick access to fmt.println. I will remove it since its not needed

render_backend_gl_linux.odin
jqcorreia: Done

window_wayland.odin
jqcorreia: I've sucessfully moved all the EGL code to the render backend. I now understand the Wayland <> EGL interplay fully. I think....

render_backend_gl.odin
```diff
@@ -132,7 +132,7 @@ gl_init :: proc(state: rawptr, window_handle: Window_Handle, swapchain_width, sw
 	}
 
 	s.ctx = ctx
-	_gl_load_procs()
+	_gl_load_procs(s.window_handle)
```
karl-zylinski: feed context, not handle

render_backend_gl_linux.odin
```diff
@@ -5,69 +5,202 @@ package karl2d
...
+	handle := (^Window_Handle_Linux)(window_handle)
+    switch whl in handle {
+    case Window_Handle_Linux_X11:
+        gl.load_up_to(3, 3, glx.SetProcAddress)
+    case Window_Handle_Linux_Wayland:
+	    gl.load_up_to(3, 3, egl.gl_set_proc_address)
+    }
+}
+
+_gl_present :: proc(window_handle: Window_Handle) {
```
karl-zylinski: feed context not handle

render_backend_gl_linux.odin
```diff
@@ -5,69 +5,202 @@ package karl2d
...
+        whl.egl_surface = egl_surface
+
+        return GL_Context_EGL { window_handle = whl^, ctx = egl_context, egl_display = egl_display }, true
+    }
+    return {}, false
 }
 
-_gl_load_procs :: proc() {
-	gl.load_up_to(3, 3, glx.SetProcAddress)
+wayland_gl_present :: proc(whl: ^Window_Handle_Linux_Wayland) {
```
karl-zylinski: this proc should tell window to just present backbuffer, not to redraw

if that can't work with wayland then perhaps we need to rethink whole organisation of these context files. Perhaps they need to be an interface with multiple implementations that are issued by the windowing ation

window_wayland.odin
```diff
@@ -0,0 +1,633 @@
...
+	s.window_handle = Window_Handle_Linux_Wayland {
+		ready = false,
+	}
+
+	display := wl.display_connect(nil)
+	s.display = display
+
+	// Get registry, add a global listener and get things started
+	// Do a roundtrip in order to get registry info and populate the wayland part of state
+	registry := wl.wl_display_get_registry(display)
```
karl-zylinski: drop all `wl_` prefixes. If you get procedure <--> type name conflict because of this, then change the wayland types to use Ada_Case

window_wayland.odin
```diff
@@ -0,0 +1,633 @@
...
+	wl.display_dispatch(s.display)
+
+	wl_callback := wl.wl_surface_frame(surface)
+	wl.wl_callback_add_listener(wl_callback, &frame_callback, nil)
+
+	egl_window := wl.egl_window_create(surface, i32(s.windowed_width), i32(s.windowed_height))
+
+	whl := &s.window_handle.(Window_Handle_Linux_Wayland) 
+	whl.redraw = true
+	whl.display = display
```
karl-zylinski: Don't store things used by window in the handle. Have it in the Wayland_State and just assign it once to the handle so the context-creator file can read it. It's only meant to exist in order to pass things from window to the file that handles context. It's a one way street. No data should be passed back using it.

window_wayland.odin
```diff
@@ -0,0 +1,633 @@
...
+	// whl.egl_display = egl_display
+	// whl.egl_config = egl_config
+	// whl.egl_surface = egl_surface
+	// whl.egl_context = egl_context
+
+	wayland_set_window_mode(init_options.window_mode)
+}
+
+wayland_shutdown :: proc() {
+	// X.DestroyWindow(s.display, s.window)
```
karl-zylinski: missing egl call

window_wayland.odin
```diff
@@ -0,0 +1,633 @@
...
+
+s: ^Wayland_State
+
+@(private="package")
+Window_Handle_Linux_Wayland :: struct {
+	ready: bool,
+	redraw: bool,
+	display: ^wl.wl_display,
+	surface: ^wl.wl_surface,
+	egl_display: egl.Display,
```
karl-zylinski: Move things that do not need to be here out of here. Either to Wayland_State or to GL_Context_EGL. Only store things in here needed to pass it to context file

window_wayland.odin
```diff
@@ -0,0 +1,633 @@
...
+		xdg_toplevel: ^wl.xdg_toplevel,
+		width: c.int32_t,
+		height: c.int32_t,
+		states: ^wl.wl_array,
+	) {
+		context = runtime.default_context()
+		// fmt.println("top level configure")
+		sw := i32(s.windowed_width)
+		sh := i32(s.windowed_height)
+		whl := s.window_handle.(Window_Handle_Linux_Wayland)
```
karl-zylinski: if you stored these things primarily in the state then you don't need to fetch window_handle ghere

window_wayland.odin
```diff
@@ -0,0 +1,633 @@
...
+	windowed_height: int,
+	events: [dynamic]Window_Event,
+	display: ^wl.wl_display,
+	compositor: ^wl.wl_compositor,
+	xdg_base: ^wl.xdg_wm_base,
+	seat: ^wl.wl_seat,
+	surface: ^wl.wl_surface,
+	toplevel: ^wl.xdg_toplevel,
+    decoration: ^wl.zxdg_decoration_manager_v1,
+	window_handle: Window_Handle_Linux,
```
karl-zylinski: Window_Handle_Linux_Wayland

render_backend_gl.odin
jqcorreia: done

render_backend_gl_linux.odin
jqcorreia: done

render_backend_gl_linux.odin
jqcorreia: Probably redraw here is just a misnomer. `redraw` in this sense is just a flag that is set by a message in the wayland protocol that is supposed to mark the buffer to be redrawn. We could, for now, just ignore redraw and go for buffer change every simulated frame

PR #24, opened 2026-01-09
Darwin support: First cut at macos support (using opengl for now)

cpoile:
Hey Karl!

First cut at adding macos support. Uses cocoa/appkit (I've named everything cocoa just for simplicity). Let me know if you'd prefer AppKit prefixes. 

I'll work on gamepad support over the weekend, then next up is Metal.

### DIFF
--- a/window_cocoa.odin
+++ b/window_cocoa.odin
@@ -68,7 +68,6 @@ cocoa_init :: proc(
 	s.events = make([dynamic]Window_Event, allocator)
 	s.width = screen_width
 	s.height = screen_height
-	s.window_mode = init_options.window_mode
 
 	// Initialize NSApplication
 	s.app = NS.Application_sharedApplication()
@@ -86,24 +85,14 @@ cocoa_init :: proc(
 	app_menu_item->setSubmenu(app_menu)
 	s.app->setAppleMenu(app_menu)
 
-	// Activate the application
-	s.app->activateIgnoringOtherApps(true)
-	s.app->finishLaunching()
-
-
-	// Create window style based on mode
-	style := NS.WindowStyleMaskTitled | NS.WindowStyleMaskClosable | NS.WindowStyleMaskMiniaturizable
-	if init_options.window_mode == .Windowed_Resizable {
-		style |= NS.WindowStyleMaskResizable
-	}
-
 	// Create the window
 	rect := NS.Rect {
 		origin = {0, 0},
 		size = {NS.Float(screen_width), NS.Float(screen_height)},
 	}
 	s.window = NS.Window_alloc()
-	s.window = s.window->initWithContentRect(rect, style, .Buffered, false)
+	s.window = s.window->initWithContentRect(rect, {}, .Buffered, false)
+	s.windowed_rect = rect
 
 	title_str := NS.String_alloc()->initWithOdinString(window_title)
 	s.window->setTitle(title_str)
@@ -111,19 +100,12 @@ cocoa_init :: proc(
 	s.window->center()
 	s.window->setAcceptsMouseMovedEvents(true)
 	s.window->makeKeyAndOrderFront(nil)
-
-	s.windowed_rect = s.window->frame()
-
-	// Now that we've recorded the starting rect, we can go fullscreen
-	if init_options.window_mode == .Borderless_Fullscreen {
-		s.window->setStyleMask({})
-		screen_frame := NS.Screen_mainScreen()->frame()
-		s.window->setFrame(screen_frame, true)
-		s.window->setLevel(.Status)
-		ce.Application_setPresentationOptions(s.app, {.HideMenuBar, .HideDock})
-	}
-
 	s.window_handle = s.window
+	cocoa_set_window_mode(init_options.window_mode)
+
+	// Activate the application
+	s.app->activateIgnoringOtherApps(true)
+	s.app->finishLaunching()
 
 	// Setup delegates for events not handled in cocoa_process_events
 	window_delegates := NS.window_delegate_register_and_alloc(
@@ -344,6 +326,11 @@ cocoa_set_window_mode :: proc(window_mode: Window_Mode) {
 		s.window->setFrame(screen_frame, true)
 		s.window->setLevel(.Normal)
 		ce.Application_setPresentationOptions(s.app, {.HideMenuBar, .HideDock})
+
+		// same as frame() b/c no decorations, but semantically more correct
+		content_rect := s.window->contentLayoutRect()
+		s.width = int(screen_frame.width)
+		s.height = int(screen_frame.height)
 	}
 }
 


karl-zylinski:
Instead of adding the mac-specific GL shader, can we use a `version 330` shader on mac and windows and then use a 300 es on webgl? So it `gl` and `webgl` default shaders, no mac default things mention in shader names

render_backend_gl_darwin.odin
```diff
@@ -0,0 +1,94 @@
...
+@private
+RTLD_DEFAULT :: rawptr(~uintptr(0) - 1) // -2 cast to pointer
+
+// the OpenGL shared library is loaded from OpenGL.framework when we initialize it in _gl_get_context
+@private
+macos_gl_set_proc_address :: proc(p: rawptr, name: cstring) {
+	(^rawptr)(p)^ = os._unix_dlsym(RTLD_DEFAULT, name)
+}
+
+_gl_present :: proc(window_handle: Window_Handle) {
```
karl-zylinski: send in the context instead and use that to flush the buffer

window_cocoa.odin
```diff
@@ -0,0 +1,440 @@
...
+	s.width = screen_width
+	s.height = screen_height
+	s.custom_context = context
+	s.window_mode = init_options.window_mode
+
+	// Initialize NSApplication
+	s.app = NS.Application_sharedApplication()
+	s.app->setActivationPolicy(.Regular)
+
+    NS.scoped_autoreleasepool()
```
karl-zylinski: Seems to have a mix of tabs and spaces for indent

window_cocoa.odin
```diff
@@ -0,0 +1,440 @@
...
+
+    menu_bar := NS.Menu_alloc()->init()
+    s.app->setMainMenu(menu_bar)
+    app_menu_item := menu_bar->addItemWithTitle(NS.AT(""), nil, NS.AT(""))
+    app_menu := NS.Menu_alloc()->init()
+    fullscreen_item := app_menu->addItemWithTitle(NS.AT("Enter Full Screen"), NS.sel_registerName(cstring("toggleFullScreen:")), NS.AT("f"))
```
karl-zylinski: Not sure we should have this. The game should decide if it wants to provide fullscreen or not I think

window_cocoa.odin
```diff
@@ -0,0 +1,440 @@
...
+				s.app->sendEvent(event)
+			}
+		} else {
+			s.app->sendEvent(event)
+		}
+	}
+
+	// Check for window resize
+    // TODO: not sure if this is where we should be updating
+	content_rect := s.window->contentLayoutRect()
```
karl-zylinski: Is there no event for when the resize happens that you could react to instead?

window_cocoa.odin
```diff
@@ -0,0 +1,440 @@
...
+		if s.window_handle.opengl_context != nil {
+			s.window_handle.opengl_context->update()
+		}
+	}
+
+	// Check for window close request
+	// (This is a simple check - a more robust solution would use a window delegate)
+	if s.window != nil && !s.window->keyWindow() {
+		// Window lost focus - send unfocused event if not already sent
+		// For now we'll skip this - would need to track focus state
```
karl-zylinski: Is there no event for focus / unfocus?

Sending unfocus events is require. Otherwise you get bugs because keys "get stuck". As a test, open the `basics` example and hold down the arrow keys to move the red dot. Try unfocusing the window while holding it and see if it stops. If it does not stop, then the unfocus event isn't sent properly. When the unfocused event happens, it will clear all the button states.

window_cocoa.odin
```diff
@@ -0,0 +1,440 @@
...
+
+	case .Windowed_Resizable:
+		if old_mode == .Borderless_Fullscreen {
+			s.window->toggleFullScreen(nil)
+		}
+		style := NS.WindowStyleMaskTitled | NS.WindowStyleMaskClosable | NS.WindowStyleMaskMiniaturizable | NS.WindowStyleMaskResizable
+		s.window->setStyleMask(style)
+
+	case .Borderless_Fullscreen:
+		s.window->toggleFullScreen(nil)
```
karl-zylinski: Is this borderless or true fullscreen? Just wondering, so the name of the mode is correct.

window_cocoa.odin
```diff
@@ -0,0 +1,440 @@
...
+	}
+}
+
+s: ^Cocoa_State
+
+@(private="package")
+Window_Handle_Darwin :: struct {
+	window:         ^NS.Window,
+	view:           ^NS.View,
+	opengl_context: ^nsgl.OpenGLContext,
```
karl-zylinski: Can't you store this in the GL_Context instead? I know you run `s.window_handle.opengl_context->update()` when window changes size. But when the window resize event, then `gl_resize_swapchain` in the backend is invoked. In there you can add a call to a new `_gl_context_viewport_resized` proc that contains the call to `s.window_handle.opengl_context->update()`.

karl-zylinski (changes requested):
Thanks for the PR. In general your PR looks very tidy and things are mostly like I want them. Good job.

Lets try to get this merged when you have the gamepads working.

Also, don't start on the Metal stuff until we have this merged. While this PR was fairly straight forward, I can't with 100% confidence say that the Metal one will be, since it might require architectural changes to the render backend interface (there might be biases that came with D3D and GL that are incompatible with Metal that we need to take into account).

window_cocoa.odin
```diff
@@ -0,0 +1,440 @@
...
+
+	is_gamepad_active = cocoa_is_gamepad_active,
+	get_gamepad_axis = cocoa_get_gamepad_axis,
+	set_gamepad_vibration = cocoa_set_gamepad_vibration,
+
+	set_internal_state = cocoa_set_internal_state,
+}
+
+import NS "core:sys/darwin/Foundation"
+import nsgl "darwin/nsgl"
```
karl-zylinski: If you manage to get the context update stuff out like I mentioned in another comment, then you can remove this import

window_cocoa.odin
```diff
@@ -0,0 +1,440 @@
...
+cocoa_set_gamepad_vibration :: proc(gamepad: int, left: f32, right: f32) {}
+
+cocoa_set_internal_state :: proc(state: rawptr) {
+	assert(state != nil)
+	s = (^Cocoa_State)(state)
+}
+
+Cocoa_State :: struct {
+	allocator:        runtime.Allocator,
+	custom_context:   runtime.Context,
```
karl-zylinski: You don't seem to need this one. It was used in other window interfaces where there was a callback in which we needed a context

render_backend_gl_darwin.odin
cpoile: I'm using a private global now. But curious, what do you mean by sending in the context? I'm not sure how we could do that and keep the nsgl out of the window dependencies...  Or do you mean make a darwin-specific version of `_gl_present`? (I wouldn't think so, that's awful :) )

window_cocoa.odin
cpoile: Ah, oops. I fixed that, I also added remove trailing whitespace, so there are some changes to other lines now in files I edited, apologies. I can revert those hunks if it's going to be a problem.

window_cocoa.odin
cpoile: Makes sense -- removed. 
Question for you -- do you want windowed mode to have the title bar and title? Or just be a bare window? (which looks kind of cool for a game)  Or maybe offer that as an option (if it's available in Win/Linux)... It does have the downside that it can't be dragged. So maybe not great.

window_cocoa.odin
cpoile: Yep, delegates. I've added them, much nicer now.

window_cocoa.odin
cpoile: Good catch, thanks. Implemented the focus/unfocus, and close wanted.

window_cocoa.odin
cpoile: Now it's borderless fullscreen. True fullscreen is not great for games (from what I understand).

window_cocoa.odin
cpoile: I added `gl_context` as a private global, rather than changing the signature just for darwin. Let me know if you have a better idea!
Nice idea for `_gl_context_viewport_resized` -- that works great.

window_cocoa.odin
cpoile: Yep, nice, thanks.

window_cocoa.odin
cpoile: Yep, removed.

cpoile:
Thanks for the great review! Made some updates, take a look when you can. 
I have a todo left, and I think I need to work on the restoring after fullscreen -> windowed a few times -- the window creeps up the screen each time, I'm probably not taking the menubar into account or something. I'll figure it out. 
And gamepads. :)

darwin/cocoa_extras/cocoa_extras.odin
```diff
@@ -0,0 +1,19 @@
+#+build darwin
+
+package cocoa_extras
+
+// Extra Cocoa/AppKit bindings not included in Odin's standard darwin Foundation bindings
```
cpoile, line 5: Not super happy about adding another package, but it's also different code than what is in the nsopengl lib.  I'd be happy putting them all in there, or here, just let me know if you'd prefer that.

render_backend_gl_darwin.odin
karl-zylinski: Can't you just do `_gl_present(s.ctx)` in the backend and have 

```
_gl_present :: proc(ctx: GL_Context) {
    handle := ctx.handle // save away handle in ctx inside `_gl_get_context`.
    // use handle
}
```

I know I wrote the intial code that passed in window_handle, but that was a bad idea since it makes people store weird stuff in the handle

window_cocoa.odin
karl-zylinski: While mix of tabs and spaces break the code's readability and makes it annoying to edit, I don't think some trailing whitespace matters. Please revert those changes.

window_cocoa.odin
karl-zylinski: It should have caption with title and title bar and all the buttons in the title bar should work properly. We can maybe do caption-less later. It's possible on win and linux too. But I will wait until someone requests it because their game needs it :)

window_cocoa.odin
karl-zylinski: True fullscreen is also useful, but it's annoying on some platforms and harder to implement, so I've named the enum `Borderless_Fullscreen` so I can add a `Fullscreen` option later when I feel like it. Borderless has the property that the resolution becomes the same as the desktop, while true fullscreen can change to any resolution.

window_cocoa.odin
karl-zylinski: Put it in the `GL_Context` state. That whole struct is unique on each platform because it is defined inside the `_darwin.odin`, `_linux.odin` etc files

darwin/cocoa_extras/cocoa_extras.odin
karl-zylinski: We can have these there for now and later do a PR into Odin and add them there. When the next release of Odin goes out we can remove this file.

window_cocoa.odin
karl-zylinski: Another reason is that reviewing becomes harder: There are now lots of changes that do not matter, so reviewing becomes way less fun. I'll continue the review once the trailing whitespace changes are reverted.

render_backend_gl_darwin.odin
karl-zylinski: I can fix it on Windows and Linux after you've changed it, if you need someone to test it.

render_backend_gl_darwin.odin
```diff
@@ -0,0 +1,92 @@
...
+@private
+RTLD_DEFAULT :: rawptr(~uintptr(0) - 1) // -2 cast to pointer
+
+// the OpenGL shared library is loaded from OpenGL.framework when we initialize it in _gl_get_context
+@private
+macos_gl_set_proc_address :: proc(p: rawptr, name: cstring) {
+	(^rawptr)(p)^ = os._unix_dlsym(RTLD_DEFAULT, name)
+}
+
+_gl_present :: proc(_: Window_Handle) {
```
karl-zylinski: `_gl_present :: proc(ctx: GL_Context)`

render_backend_gl_darwin.odin
```diff
@@ -0,0 +1,92 @@
...
+	gl_context->flushBuffer()
+}
+
+_gl_context_viewport_resized :: proc(_: Window_Handle) {
```
karl-zylinski: `_gl_context_viewport_resized :: proc(ctx: GL_Context)`

karl-zylinski:
> Thanks for the great review! Made some updates, take a look when you can. I have a todo left, and I think I need to work on the restoring after fullscreen -> windowed a few times -- the window creeps up the screen each time, I'm probably not taking the menubar into account or something. I'll figure it out. And gamepads. :)

Yes, I had similar issues with "creeping" windows on Windows. There I had to set the windowed position on a rectangle and then push the rectangle through something called `win32.AdjustWindowRectblabla` in order to make it a adapt it the rectangle to the caption etc.

render_backend_gl_darwin.odin
```diff
@@ -0,0 +1,92 @@
...
+import "darwin/nsgl"
+import "log"
+
+GL_Context :: ^nsgl.OpenGLContext
+
+@private
+gl_context: ^nsgl.OpenGLContext
+
+_gl_get_context :: proc(window_handle: Window_Handle) -> (GL_Context, bool) {
+	using nsgl
```
karl-zylinski: Remove `using nsgl`. It's not good for parameters and variables.

window_cocoa.odin
```diff
@@ -0,0 +1,449 @@
...
+	s.window->setTitle(title_str)
+
+	s.window->center()
+	s.window->setAcceptsMouseMovedEvents(true)
+	s.window->makeKeyAndOrderFront(nil)
+
+	s.windowed_rect = s.window->frame()
+
+	// Now that we've recorded the starting rect, we can go fullscreen
+	if init_options.window_mode == .Borderless_Fullscreen {
```
karl-zylinski: Perhaps you can just skip setting `s.window_mode` and call `cocoa_set_window_mode(window_mode)` here instead? Or perhaps at the end of init. See how the win32 windowing does it.

If it doesn't work, then that is fine.

karl-zylinski (changes requested):
Thanks for the changes. It looks quite good. If we can just fix the things about how the context vs window handle is passed, then this would be good to merge. If you don't understand what all that means, then let me know and I can jump into the branch and do it.

(the gamepad stuff can come separately if you don't have time for it right now)

render_backend_gl_darwin.odin
cpoile: thanks -- I understood but had different field names. :)  
naming is the worst, so I'll do whatever you want with that (now and in the future)

cpoile:
> There I had to set the windowed position on a rectangle and then push the rectangle through something called win32.AdjustWindowRectblabla in order to make it a adapt it the rectangle to the caption etc.

The solution on mac is to fix the order of operations -- add the title bar back and /then/ restore the window position.

window_cocoa.odin
cpoile: Yep, works nicely, thanks.

karl-zylinski:
Merged. Thanks for the PR!

PR #41, opened 2026-01-16
Linux gamepad support

jqcorreia:
Leverage `joydev`  kernel module to implement a primitive gamepad support in Linux.
The gamepad device files are present at `/dev/input/js*` and are simply a stream of events that can be trivially transformed into axis information.

### DIFF
--- a/platform_linux.odin
+++ b/platform_linux.odin
@@ -78,45 +78,7 @@ linux_init :: proc(
 		allocator,
 	)
 
-	// Create gamepads that were connected on program boot
-	{
-		// Gamepads are described by device files at path `/dev/input/eventXX`
-		devices_handle, devices_handle_ok := os.open("/dev/input") 
-
-		if devices_handle_ok != nil {
-			return 
-		}
-
-		defer os.close(devices_handle)
-
-		file_infos, file_infos_ok := os.read_dir(devices_handle, -1, frame_allocator)
-
-		if file_infos_ok != nil {
-			return 
-		}
-
-		gamepad_idx := 0
-
-		for fi in file_infos {
-			if !strings.starts_with(fi.name, "event") {
-				continue
-			}
-
-			if !evdev.is_device_gamepad(fi.fullpath) {
-				continue
-			}
-
-			if gamepad_idx >= MAX_GAMEPADS {
-				log.errorf("A maximum of %v gamepads is supported", MAX_GAMEPADS)
-				break					
-			}
-
-			if gamepad, gamepad_ok := linux_create_gamepad(fi.fullpath); gamepad_ok {
-				s.gamepads[gamepad_idx] = gamepad
-				gamepad_idx += 1
-			}
-		}
-	}
+	linux_create_connected_gamepads()
 
 	// Set up monitoring for new gamepads. This uses udev. It looks for `input` devices being added.
 	// The monitor is polled in `linux_get_events`.
@@ -207,6 +169,45 @@ linux_set_size :: proc(w, h: int) {
 
 linux_get_window_scale :: proc() -> f32 {
 	return s.win.get_window_scale()
+}
+
+linux_create_connected_gamepads :: proc() {
+	// Gamepads are described by device files at path `/dev/input/eventXX`
+	devices_handle, devices_handle_ok := os.open("/dev/input") 
+
+	if devices_handle_ok != nil {
+		return 
+	}
+
+	defer os.close(devices_handle)
+
+	file_infos, file_infos_ok := os.read_dir(devices_handle, -1, frame_allocator)
+
+	if file_infos_ok != nil {
+		return 
+	}
+
+	gamepad_idx := 0
+
+	for fi in file_infos {
+		if !strings.starts_with(fi.name, "event") {
+			continue
+		}
+
+		if !evdev.is_device_gamepad(fi.fullpath) {
+			continue
+		}
+
+		if gamepad_idx >= MAX_GAMEPADS {
+			log.errorf("A maximum of %v gamepads is supported", MAX_GAMEPADS)
+			break					
+		}
+
+		if gamepad, gamepad_ok := linux_create_gamepad(fi.fullpath); gamepad_ok {
+			s.gamepads[gamepad_idx] = gamepad
+			gamepad_idx += 1
+		}
+	}
 }
 
 linux_create_gamepad :: proc(device_path: string) -> (Linux_Gamepad, bool) {


karl-zylinski:
> Still lacking button support and we should think about a gamepad update loop/proc since right now i'm piggybacking on `linux_gamepad_get_axis()`.

Try doing it in `get_events` instead. That is called once per frame before any input state is updated.

platform_linux.odin
```diff
@@ -70,6 +70,8 @@ linux_init :: proc(
 		options,
 		allocator,
 	)
+
+    s.gamepad_controller = gamepad_new_controller()
```
karl-zylinski: Weird name because `controller` is a synonym for gamepad. So it looks like `gamepad_gamepad` :)

gamepad_linux.odin
```diff
@@ -0,0 +1,432 @@
...
+foreign _udev {
+    @(link_prefix = "")
+	udev_new :: proc() -> ^udev ---
+	unref :: proc(_: ^udev) ---
+
+	device_get_devnode :: proc(_: ^udev_device) -> cstring ---
+	device_get_action :: proc(_: ^udev_device) -> cstring ---
+	device_unref :: proc(_: ^udev_device) ---
+
+	monitor_new_from_netlink :: proc(_: ^udev, _: cstring) -> ^udev_monitor ---
```
karl-zylinski: don't skip param names in bindings: It's hard to know what this `cstring` is

gamepad_linux.odin
```diff
@@ -0,0 +1,432 @@
...
+import "core:strings"
+import "core:sys/linux"
+import "core:sys/posix"
+
+// udev minimal bindings for gamepad connection listening
+udev :: struct {}
+udev_device :: struct {}
+udev_monitor :: struct {}
+
+foreign import _udev "system:udev"
```
karl-zylinski: udev_lib

gamepad_linux.odin
```diff
@@ -0,0 +1,432 @@
...
+    gamepads : []Linux_Gamepad,
+
+    udev: ^udev,
+    udev_mon: ^udev_monitor,
+    udev_fd: i32,
+}
+
+gamepad_new_controller :: proc(max_gamepads: int = 4) -> Linux_Gamepad_Controller {
+    // Initialize present devices
+    gamepads := make([]Linux_Gamepad, 4)
```
karl-zylinski: Use fixed size array -- This code is over generalised. We need to solve the exact problem we have. Also I don't think the Linux_Gamepad_Controller is needed. You can just have a fixed array of gamepads in the `Linux_State` struct in `Platform_Linux`.

In fact, you can probably merge all of this file into `Platform_Linux` and simplify things by breaking down the barriers between them

gamepad_linux.odin
```diff
@@ -0,0 +1,432 @@
...
+                return pad, true
+            }
+        }
+    }
+
+    return Linux_Gamepad{}, false
+}
+
+gamepad_poll :: proc(gamepad: ^Linux_Gamepad) -> []Linux_GamepadEvent {
+	res: [dynamic]Linux_GamepadEvent
```
karl-zylinski: Memory leak of this array every time you use this proc

If you combine the gamepad and platform files then you can probably add straight to the standard event array that exists in `platform_linux.odin`

karl-zylinski:
Was there plan on supporting gamepad vibration?

platform_linux.odin
jqcorreia: At the last time I decided in a controller because of having the file descriptors for the udev monitor. But I suppose we can have them on the state no problem.

gamepad_linux.odin
jqcorreia: I've separated the code trying to keep the platform code readable and to-the-point, without all the udev bindings and evdev consts, enums and unions. But it can be easily moved to the platform file.

jqcorreia:
I've moved udev and evdev code to packages inside /linux where I think it makes sense to live. Refactored in order to get rid of the controller and gamepad_linux.odin altogether.

jqcorreia:
> Was there plan on supporting gamepad vibration?

I've implemented it on 8a3c6feb54689d0c43a1a5a8776fa81c38df0ed2

linux/evdev/evdev.odin
```diff
@@ -0,0 +1,233 @@
...
+	HAT3Y      = 0x17,
+	PRESSURE   = 0x18,
+	DISTANCE   = 0x19,
+	TILT_X     = 0x1a,
+	TILT_Y     = 0x1b,
+	TOOL_WIDTH = 0x1c,
+}
+
+FF_Effects :: enum {
+	FF_RUMBLE   = 0x50,
```
karl-zylinski: drop FF_ prefix on enum members

linux/evdev/evdev.odin
```diff
@@ -0,0 +1,233 @@
...
+
+// This is the first and base button
+// This is used as the bit index to check for existence
+// and also the as the event code
+BTN_GAMEPAD :: 0x130
+
+// In linux/input.h there are 15 different mapped buttons + d-pad 
+// that for some reason appear further down....
+Button :: enum u32 {
+	BTN_A          = BTN_GAMEPAD,
```
karl-zylinski: drop BTN prefix on enum members

platform_linux.odin
```diff
@@ -108,15 +147,319 @@ linux_get_window_scale :: proc() -> f32 {
...
+				trigger = evdev.ff_trigger{button = 0, interval = 0},
+				replay = evdev.ff_replay{length = 0, delay = 0},
+			}
+			effect.u.rumble = evdev.ff_rumble_effect {
+				strong_magnitude = 0,
+				weak_magnitude   = 0,
+			}
+			linux.ioctl(linux.Fd(fd), evdev.EVIOCSFF(), cast(uintptr)&effect)
+			gamepad.rumble_effect_id = u32(effect.id)
+            gamepad.has_rumble_support = true
```
karl-zylinski: tabs

linux/evdev/evdev.odin
jqcorreia: done

linux/evdev/evdev.odin
jqcorreia: done

platform_linux.odin
jqcorreia: done, i'm still tuning my setup to work with non-spaces projects. sorry about that.

karl-zylinski: @jqcorreia I recommend this setting in the github settings -> "appearance"
image

It will make tabs vs spaces stand out much more. You just enable it and make sure to look through the Changed files and you will see any tabs that are wrong.

linux/evdev/evdev.odin
```diff
@@ -0,0 +1,233 @@
...
+
+	return(
+		((dir) << _IOC_DIRSHIFT) |
+		((type) << _IOC_TYPESHIFT) |
+		((nr) << _IOC_NRSHIFT) |
+		((size) << _IOC_SIZESHIFT) \
+	)
+}
+
+// This is not working, not sure why size_of(T) is 8 when i pass it input_absinfo
```
karl-zylinski: `reflect.size_of_typeid(T)`

+

`import "core:reflect"`

It is a funny corner case, the size you're getting when you do `size_of(T)` is the size of a typeid. It's a parapoly funnybusiness.

jqcorreia:
> @jqcorreia I tested it now.
> 
> Only broken thing is that triggers go between -1 and 1, it should go between 0 and 1 (different from the sticks which go from -1 to 1)
> 
> Also, I couldn't compile unless I installed `systemd-devel`, because it couldn't find `libudev` otherwise. But I guess that's expected?

I'll take a look at the ranges, i'm using the range declared by the device itself, and I believe k2d is expecting 0,1?

Regarding udev dependency, yes, it is needed for the hotplug mechanism, without it it would be way tougher to get device connect/disconnect monitoring. But needing an extra package (a -devel one) on a typical linux install is strange. I'll check what's the lib that it is tryink to link to.

karl-zylinski:
@jqcorreia I think I am done with my review and my changes now. Does it look OK? Have I done anything inadvisable when it comes to the linux stuff?

karl-zylinski:
Thank you for the PR!! All that evdev stuff would have given me a headache :) Great stuff

Issue #30, opened 2026-01-11
Mac gamepad support


karl-zylinski:
There is mac window support, but it is missing support for gamepads

cpoile:
Hey Karl, I can take it -- I should be able to finish it this weekend.

karl-zylinski:
Fixed in #44

PR #44, opened 2026-01-17
Controller & haptics support for mac #30

cpoile:
#### Summary

- Adds gamecontroller support for mac. 
- At the moment we only support modern controllers through the GameController framework. I've tested with an XBox controller, but that's the only modern controller I have.  
- My switch 1 Joy-Cons don't register, unfortunately. But Steam doesn't seem to support them either.
- My PS3 DualShock 3's do not work either. I tried to support those, but it requires IOKit. We would need to use a lot of custom code to support each controller. If we want to do that, I would recommend copying from glfw's simpler version.  I tried to convert some of it to Odin and it was a real pain and didn't end up working, so I reverted the attempt. Another option is copying SDL3's approach, but that's even more code. Something to think about if we want to support older controllers.

#### Issue
- Solves #30

### DIFF
--- a/platform_mac.odin
+++ b/platform_mac.odin
@@ -628,7 +628,7 @@ controller_is_registered :: proc(controller: ^gc.Controller) -> bool {
 }
 
 remove_controller :: proc(controller: ^gc.Controller) {
-	for &gamepad, gamepad_index in s.gamepads {
+	for &gamepad in s.gamepads {
 		if gamepad.controller == controller {
 			// haptic support is only available in 11.0.0
 			when ODIN_MINIMUM_OS_VERSION >= 11_00_00 {


platform_mac.odin
```diff
@@ -247,6 +288,22 @@ mac_get_events :: proc(events: ^[dynamic]Event) {
 		}
 	}
 
+	// Poll gamepad buttons and generate events
```
cpoile, line 293: Alternatively we can use the value change handler (callback) system. But it's more complicated, and SDL3 and glfw both use polling, so that's the route I took here.

darwin/gamecontroller/corehaptics.odin
```diff
@@ -0,0 +1,165 @@
+#+build darwin
+
+package gamecontroller
```
karl-zylinski: Maybe use a less generic package name. "karl2d_darwin_gamecontroller" for example good. package names can after all collide (don't change the folder name however :))

platform_mac.odin
karl-zylinski: I think polling is fine. It's just `4*max(Gamepad_Button)` iterations, it's not much.

karl-zylinski (commented):
Thank you for another very high quality Pull Request!! ❤️ 

I have almost no feedback to give.

I will boot up my ancient macbook and give this a go using an xbox and ps4 controller. I know PS3 didn't work, it doesn't on Windows either (unless you install wrapper drivers). PS4 and PS5 are different however, so I'll see if it works out of the box or not.

darwin/gamecontroller/corehaptics.odin
cpoile: Hah, nice, didn't realize that could cause naming collisions. Thanks!

karl-zylinski: It's the name the package uses at link time. Think of it as a prefix that gets appended to all the procs etc in your package. At link time all the things in all packages just form a flat hierarchy with these prefixes to distinguish things. If something else in the same translation unit (another package used in same program) uses the same package name, then the linker will cry.

karl-zylinski:
@moomerman You're a mac person I've come to understand. Ignore this if you don't have time. But do you happen to have a gamepad and want to try if this branch works for you? I.e. hop on @cpoile's branch and run the `examples/gamepad` example and see if everything works, including vibration when you hold down the left/right trigger.

cpoile:
Ok, I think I've got conditional compilation working -- can you check?  

About unplugging: I was able to test with my ps3 controller (the buttons don't work, but it does show up), and
1) unplugging does not change the plugged in controller's position.  But,
2) plugging in a new controller (xbox) will take spot 1 (at least on my machine). I don't know the rules mac is using, but it seems it prioritizes that over the ps3 controller.

Is that behavior consistent with windows? If it's not, I'll have to add the bookkeeping code.

platform_mac.odin
```diff
@@ -40,6 +44,25 @@ Mac_State :: struct {
...
+}
+
+when ODIN_MINIMUM_OS_VERSION >= 11_00_00 {
+	Gamepad_Haptics :: struct {
+		haptic_engine_left_right: [2]^gc.HapticEngine,
+		haptic_player_left_right: [2]^gc.HapticPatternPlayer,
+		old_intensity_left_right: [2]f32,
+	}
+
+	gamepad_haptics: [MAX_GAMEPADS]Gamepad_Haptics
```
karl-zylinski: You can't keep global state outside of the `Mac_State` struct. That won't work with hot reloading.

Regarding the "haptics gating": You don't need to gate the types. Those don't bind to any code in the library. You don't need the gating at all in corehaptics.odin. The only gating you need is `when` blocks around the usages of the procs. The only `when` you need is the one in the `mac_set_gamepad_vibration`, `remove_controller` and the`when` that encloses `stop_haptic_player`, `init_haptic_engine` and `create_haptic_player`.

How this works is that the foreign import of corehaptics only happens if there is a usage of a proc in that library. If all the proc calls are remove using `when`, then the library isn't even loaded.

This means that you can put the gamepad_haptics stuff straight into the gamepad type, remove the gating in the haptics binding file and then its all OK.

platform_mac.odin
```diff
@@ -103,6 +126,31 @@ mac_init :: proc(
...
+	// Add already connected controllers
+	poll_for_new_controllers()
+
+	// Setup listeners for connected/disconnected controllers
+	notificationCenter := NS.NotificationCenter_defaultCenter()
+
+	notificationCenter->addObserverForName(
+		gc.DidConnectNotification, nil, nil,
+		NS.Block_createGlobalWithParam(s, proc "c" (s: rawptr, n: ^NS.Notification) {
+			context = runtime.default_context()
```
karl-zylinski: Do not use default context. It does not have loggers or the correct allocators set up. A better idea is to save away `s.odin_ctx = context` in `mac_init` and then use that value here.

karl-zylinski (commented):
The book keeping changes look good to me. But I can't test it due to my old mac.

I put in a comment about the haptics gating and how to simplify it.

Other than that it looks good to me, if you and @moomerman figure out that it works fine, then I say that this is good to merge!

Just for clarity, how I expect these things to work is this:
- Start `examples/gamepad`
- Any controller (up to 4) get set up properly.
- If controller 1, 2 and 3 are all connected and you disconnect 2, then the remaining ones stay as controller 1 and 3 (this is important because in the gameplay code the gamedevs will save the Gamepad_Index as the controller for a player in a multiplayer game. It should stay that way until the controller is unplugged!)
- If you again connect a new controller and there is only a controller 1 and 3, then the new one ends up as controller 2.

(the controller index is actually numbered 0, 1, 2 and 3, but used the 1-based indices here because that is what is shown in the gamepad example...)

karl-zylinski:
By the way, and I have no idea if this is important, but I see that you have this proc bound in the bindings: https://developer.apple.com/documentation/gamecontroller/gccontroller/startwirelesscontrollerdiscovery(completionhandler:)?language=objc

Should it be used? Is it for pairing wireless controllers or for simply finding them so they become usable from the game? Feel free to ignore this if it is irrelevant. I'm just curious. The docs are quite vague.

platform_mac.odin
cpoile: Ah, I understand better now. I was using `@require` which was forcing the linking. Removing that fixed it. Thanks!
I tested setting the minimum-os-version and then inspecting with otool, and we're good -- no linking for 10.15.0.

platform_mac.odin
cpoile: Gotcha, thanks.

karl-zylinski:
Alright. Given that testing shows that things work like I described in my "Just for clarity" list earlier, then we are good to go! Let me know if it all works and I'll merge this.

moomerman:
OK I think it is all working as expected 🎉  I think I may have inadvertently been losing focus on the game window in my other test, I just tried it again ensuring I had focus and it worked great!  Disconnecting and reconnecting all worked as expected.  Nice work @cpoile!

PR #52, opened 2026-01-23
Add an option to embed wasm and js files into singular html

mfbulut:
I’m not sure how useful this feature will be for others, but I personally like using single-file html builds for small cross platform games. You don't even need to have a http server just click and play

Embedding everything does increase the file size quite a bit
It uses base64 for wasm so 2 out of every 8 byte is wasted 

* Normal: 1.5 MB
* Single file: 2 MB

But when zipped the size difference is smaller:

* Normal: 750 KB
* Single file: 900 KB


karl-zylinski:
Thanks for the PR.

For web builds you probably don't want this, as it will make the WASM file reload more often than necessary.

I understand the need for local builds (no cross-origin errors), but I'm not sure this is the correct approach.

PR #54, opened 2026-01-23
Added Rotation and Origin to draw_text_ex()

DallinTG:
This is a very small change i am just using the the rot and origin from the preexisting draw_texture_ex() just had to move some numbers around so that it properly used the origin .the rot just gets passed through


karl-zylinski:
Thanks for the PR. I borrowed your idea of how to do the rotation as part of a major drawing refactor I am doing: https://github.com/karl-zylinski/karl2d/pull/121

PR #58, opened 2026-01-26
Add keys to chars & getting the chars just pressed.

blob1807:
Add the ability to go from `Keyboard_Key` to `rune`/grapheme for the current keyboard and getting the `rune`s of the keys just pressed.
It's currently just for Windows, but before I go adding the other platforms. I'd would ask that it be briefly reviewed. 
This started as a personal thing for a text box, and I want to know if it's on the right track or even wanted.

If I don't end up continuing this. Future PRs are okay to reference & use this PR, in whole or parts.

### DIFF
--- a/karl2d.doc.odin
+++ b/karl2d.doc.odin
@@ -185,6 +185,22 @@ key_went_up :: proc(key: Keyboard_Key) -> bool
 // Returns true if a keyboard is currently being held down. Set when 'process_events' runs.
 key_is_held :: proc(key: Keyboard_Key) -> bool
 
+// Returns the rune for the given key. If there's more then one, the first is returned.
+// Repects keyboard layout.
+key_to_rune :: proc(key: Keyboard_Key, shift := false) -> rune
+
+// Returns a cloned array of the runes for given key's grapheme.
+// Repects keyboard layout.
+key_to_grapheme :: proc(key: Keyboard_Key, alloc: runtime.Allocator, shift := false) -> []rune
+
+// Returns cloned array of the runes of the keyboard keys that went down between the current and the previous frame.
+// Set when 'process_events' runs.
+get_key_runes :: proc(alloc: runtime.Allocator) -> []rune
+
+// Returns cloned string of the runes of the keyboard keys that went down between the current and the previous frame.
+// Set when 'process_events' runs.
+get_key_runes_as_string :: proc(alloc: runtime.Allocator) -> string
+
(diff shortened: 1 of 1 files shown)


karl2d.odin
```diff
@@ -492,6 +503,36 @@ key_is_held :: proc(key: Keyboard_Key) -> bool {
...
+    return s.platform.keyboard_key_to_grapheme(key, alloc, shift)
+}
+
+// Returns cloned array of the runes of the keyboard keys that went down between the current and the previous frame.
+// Set when 'process_events' runs.
+get_key_runes :: proc(alloc: runtime.Allocator) -> []rune {
+	if len(s.key_went_down_runes) == 0 {
+		return nil
+	}
+	return slice.clone(s.key_went_down_runes[:], alloc)
```
karl-zylinski, line 524: I would not allocate in this proc and instead just return the slice and in the docs write that it's only valid this frame. This is similar to how `get_events` is only valid this frame. If the users wants a copy of the slice then they can clone it themselves.

karl2d.odin
```diff
@@ -1921,6 +1962,7 @@ State :: struct {
 	mouse_wheel_delta: f32,
 
 	key_went_down: #sparse [Keyboard_Key]bool,
+	key_went_down_runes: [dynamic]rune,
```
karl-zylinski, line 1965: Not sure about the use of `went_down` here since there is no `went_up` to go with it. Perhaps it's `typed_runes` or `written_runes` or something, I'm not sure :)

karl2d.odin
```diff
@@ -492,6 +503,36 @@ key_is_held :: proc(key: Keyboard_Key) -> bool {
 	return s.key_is_held[key]
 }
 
+// Returns the rune for the given key. If there's more then one, the first is returned.
+// Repects keyboard layout.
+key_to_rune :: proc(key: Keyboard_Key, shift := false) -> rune {
```
karl-zylinski, line 508: I'm not 100% sure what the use case for this is. If you type keys then you get them from `get_keys_rune` or `get_keys_grapheme`. When would you need to go the other way?

blob1807: For displaying the actual character on the keyboard, for things like controls.  Though `key_to_grapheme` would be enough for that.
Not so much intended to be used instead of `get_key_runes`.

karl2d.odin
blob1807: Sorry about that, it's a left over from when I wanted a `went_up` stream too. I do like `written_runes` better.

PR #59, opened 2026-01-26
Mac Memory Leak Fixes

### LINKED ISSUE #57: Mac: Memory leak
Running the memory tracker results in this for a very simple karl2d program that just performs a clear:

```
/odin/core/sys/darwin/Foundation/NSBlock.odin(96:9) leaked: 48 bytes
/karl2d/platform_mac.odin(191:26) leaked: 16 bytes
/odin/core/sys/darwin/Foundation/NSBlock.odin(96:9) leaked: 48 bytes
```

EliasPrescott:
Fixes #57

Commit ce9cf51 fixes the small leak in `platform_mac.odin` by freeing the glue state struct.

And cc750d5 handles the NSBlock leaks by allocating them with a dynamic arena. I am definitely open to a different approach here, but an arena felt nice because it prevents us from having to store each block pointer in the state just to free them later.

### DIFF
--- a/platform_mac.odin
+++ b/platform_mac.odin
@@ -1,4 +1,5 @@
 #+build darwin
+#+vet explicit-allocators
 #+private file
 
 package karl2d
@@ -47,6 +48,8 @@ Mac_State :: struct {
 	window_render_glue: Window_Render_Glue,
 
 	gamepads:           [MAX_GAMEPADS]Gamepad,
+	gc_connect_blk:     ^NS.Block,
+	gc_disconnect_blk:  ^NS.Block,
 }
 
 Gamepad :: struct {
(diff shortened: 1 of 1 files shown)


platform_mac.odin
```diff
@@ -47,6 +48,7 @@ Mac_State :: struct {
 	window_render_glue: Window_Render_Glue,
 
 	gamepads:           [MAX_GAMEPADS]Gamepad,
+	block_arena:		mem.Dynamic_Arena,
```
karl-zylinski: don't use tabs for alignment: Since there are chars before on the line it will not align like one expects on different tab sizes. Essentially: Tab for indentation, spaces for alignment. Picture to show how it looks for me (I have github set to 8 spaces to easily see tab vs spaces errors)

Image

EliasPrescott: Okay thanks. Tabs are new for me so I'm still getting used to them 😄

PR #62, opened 2026-01-28
Fix for mac_set_screen_size

cpoile:
#### Summary

- For mac we should be setting the content size (which doesn't include the decorations), not changing the frame size (which does include the decorations).
- For example, if you called `k2.set_screen_size(500, 400)`, you would actually only get (500, 378) or so.


karl-zylinski:
Excellent. I love when libraries pay attention to this stuff.

PR #71, opened 2026-01-31
Mac: Clear mouse state on windowDidEndLiveResize event

### LINKED ISSUE #31: Mac: Mouse click hangs if you resize window
`odin run tests/karl_playground`

Try resize window.

Note how the cat picture keeps spinning after window resize. It should only spin while mouse is held down.

When you finish resizing a `mouse down` event is sent. But there is never any mouse up event. So it gets stuck. Not sure if some mouse capturing is missing or if we just need to clear the mouse button state after a resize.

moomerman:
Hook into the `windowDidEndLiveResize` event and emit the `Event_Window_Unfocused` event which resets the state of the mouse (and other inputs).

Wasn't sure if we want to have another specific event for this, the `Event_Window_Unfocsed` event currently does what we want.

https://github.com/user-attachments/assets/bf4cd70b-0570-4116-841d-b9371a3c130b

Fixes #31


karl-zylinski:
Thanks for the PR. Yes, it is a bit weird that it uses the Unfocused event.

An idea that may or may not work: Check for the liveResizeStart event and if you are in a live resize, then only send the `Event_Screen_Resize` event when the live resize ends. This should also cause less backbuffer resize spam. But perhaps that doesn't solve the mouse button getting stuck? Perhaps we need a `Window_Event_Reset_Input` or something like that. I'm not 100% sure

karl-zylinski:
I changed the implementation and merged this. Thanks for your PR

PR #73, opened 2026-02-01
Fix wayland mouse button mapping

henbr:
Fixes all mouse buttons reported as left button under Wayland. 

Links:
https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_pointer-event-button
https://github.com/torvalds/linux/blob/162b42445b585cd89f45475848845db353539605/include/uapi/linux/input-event-codes.h#L357


karl-zylinski:
Thanks for the PR!

PR #74, opened 2026-02-02
Fix incorrect input format for OpenGL shaders when passing layout_formats

lucaspoffo:
When calculating the stride the OpenGL renderer ignores the layout_formats passed by the user.


karl-zylinski:
Thanks for the PR

Issue #25, opened 2026-01-09
Support mouse cursor showing/hiding


iddev5:
Use case: custom cursor for games, hiding cursor if the game doesn't require it

As per my knowledge, this is supported on all desktop platforms including Linux using XCursor

This should be implemented with two functions, so that it is configurable at runtime:
- k2.set_cursor_shown(bool)
- k2.is_cursor_shown()

(Function names for example only)

PR #76, opened 2026-02-07
Support mouse cursor show/hide

moomerman:
This adds:

```
k2.set_cursor_visible(bool)
```

Current state of implementation:

- [x] mac
- [x] web
- [x] windows
- [x]  linux

## Notes

On macOS we need to do more work to track the state of the mouse when it enters/leaves the window.  I did that as a [separate commit](https://github.com/karl-zylinski/karl2d/pull/76/commits/ff5865c28f31bb70f1ba46b146560b5801901bdb) in case there's a better way.

Hiding the cursor on `examples/mouse` to demonstrate the feature.

ref #25


DerTee:
Are you sure that it's safe to assume that no other application changes the visibility of the mouse cursor in the `get`-procs and you can just take the internal state that is purely determined by the procedure calls in karl2d? I would have expected that platform native procedures would have to be called to ensure that.

I'd be happy to help you with the implementation, if you want that.

# Web
Your current implementation on Web completely relies on the internal state being true, but I'd be very surprised if it wasn't possible for some random JS or browser plugin running on the same page to change the cursor visibility.

# Windows
On Windows the simple `get` proc is not enough, see https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-showcursor and specifically this bit:

> Windows 8: Call [GetCursorInfo](https://learn.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-getcursorinfo) to determine the cursor visibility.

About ShowCursor on Windows:
> This function sets an internal display counter that determines whether the cursor should be displayed. The cursor is displayed only if the display count is greater than or equal to 0. If a mouse is installed, the initial display count is 0. If no mouse is installed, the display count is –1.

moomerman:
I've simplified this PR to just `set_cursor_visible(bool)` and removed any state management.

karl-zylinski:
I'm looking into fixing the linux part now

karl-zylinski:
I have implemented the Wayland and X11 things now. On wayland I don't support unhiding without re-entering window for now, but that may be fixed later when we do some other cursor things.

The mac stuff looks fine to me, if you're happy with it then we can merge this.

karl-zylinski:
Thank you for your contribution

PR #83, opened 2026-02-16
add hardware cursors

brunodeangelis:
On Linux Wayland, on cursor destroy, for older compositors (<2023) the cursor is gonna be invisible until it exits the window and comes back. This is because we use [Cursor shape](https://wayland.app/protocols/cursor-shape-v1#wp_cursor_shape_manager_v1) (also see https://gitlab.freedesktop.org/wayland/wayland-protocols)


brunodeangelis (commented):
-

karl2d.odin
```diff
@@ -1795,6 +1800,56 @@ get_default_font :: proc() -> Font {
...
 
+//---------//
+// CURSORS //
+//--------//
+
+create_cursor :: proc(pixels: []Color, width: int, height: int, hotspot: [2]int) -> Cursor {
+	if len(pixels) != width * height {
+		log.error("You tried creating a cursor but the pixels buffer has uneven length to the size of the image.")
+		return -1 // What would be best to return here?
+	}
```
brunodeangelis: @karl-zylinski any comments on the return value here?

karl-zylinski: Use `0` to mean no cursor (default cursor). Push a dummy element at index zero so that slot is never used.

karl2d.odin
```diff
@@ -1967,6 +1975,58 @@ get_default_font :: proc() -> Font {
 	return s.default_font
 }
 
+//---------//
+// CURSORS //
+//--------//
+
+create_cursor :: proc(pixels: []Color, width: int, height: int, hotspot: [2]int) -> Cursor {
+	if len(pixels) != width * height {
+		log.error("You tried creating a cursor but the pixels buffer has uneven length to the size of the image.")
```
karl-zylinski: This wording is weird. Should probably be something about how the pixels buffer should have length width*height. You could even add in the current and expected lengths

karl2d.odin
```diff
@@ -1967,6 +1975,58 @@ get_default_font :: proc() -> Font {
...
+		return 0
+	}
+
+	// We make our own copy of the pixels because the OS might not make its own (like macOS),
+	// in which case it will keep referencing the ones we sent.
+	new_pixels := make([]Color, width * height, s.allocator)
```
karl-zylinski: slice.clone(pixels, s.allocator)

karl2d.odin
```diff
@@ -1967,6 +1975,58 @@ get_default_font :: proc() -> Font {
...
+	copy(new_pixels, pixels)
+
+	idx := len(s.cursors)
```
karl-zylinski: it's a bit weird that the length of the cursors array keeps growing and growing. We never reuse any old indices. Perhaps you can use a handle map instead, or add a freelist? Not sure if it really matters though, since games may not make that many cursors. Or perhaps some do?

platform_linux_window_wayland.odin
```diff
@@ -494,11 +522,71 @@ wl_set_window_mode :: proc(window_mode: Window_Mode) {
 	}
 }
 
+wl_create_cursor :: proc(pixels: []Color, width: int, height: int, hotspot: [2]int) -> Cursor_Data {
+	stride := width * 4
+	size := stride * height
+
+	fd, _ := linux.memfd_create("cursor", {})
```
karl-zylinski: Chek the errors. Log an error and return {} if one happens

platform_windows.odin
```diff
@@ -225,19 +229,19 @@ windows_get_style :: proc(window_mode: Window_Mode) -> win32.DWORD {
 	switch window_mode {
 	case .Windowed:
 		style = win32.WS_OVERLAPPED |
-	            win32.WS_CAPTION |
```
karl-zylinski, line 290: You have changed random indent here.

examples/cursors/gauntlet.png
karl-zylinski, line 1: Is there some license stuff for using this file?

examples/cursors/pointer.png
karl-zylinski, line 1: according to my github diff, this is the same file as default.png. Is that intentional?

karl-zylinski (changes requested):
I looked at this and it looks quite good. Great work.

In addition to my comments: What do we do for web builds?

Now I'll give the code a run on some platforms.

examples/cursors/pointer.png
karl-zylinski: Nevermind, I didn't see the difference haha

platform_linux_window_x11.odin
```diff
@@ -300,6 +303,41 @@ x11_set_window_mode :: proc(window_mode: Window_Mode) {
 	}
 }
 
+x11_create_cursor :: proc(pixels: []Color, width: int, height: int, hotspot: [2]int) -> Cursor_Data {
+	for i in 0 ..< len(pixels) {
+		r := pixels[i].r
+		g := pixels[i].g
+		b := pixels[i].b
+		a := pixels[i].a
+		pixels[i] = {r, g, b, a}
```
karl-zylinski: This line should be `pixels[i] = {b, g, r, a}`

platform_linux_window_x11.odin
karl-zylinski, line 1: The X11 cursors work fine on my end, even under xwayland.

examples/cursors/gauntlet.png
brunodeangelis: Public Domain as per: https://www.rw-designer.com/cursor-set/gauntlet-medieval

karl2d.odin
brunodeangelis: I agree and pondered over this, I kinda copied what `Font`s do. A handle map felt to me a bit heavy handed for this, as yeah I personally don't expect many cursors that are created and destroyed. My personal use case is to create perhaps 8 cursors at most and never delete them, use the same 8 always. But can't say the same for other people or games obviously. What do you think?

platform_linux_window_x11.odin
brunodeangelis: I restarted my machine and it worked then. Seems it was buggy on my end!

brunodeangelis:
> I looked at this and it looks quite good. Great work.
> 
> In addition to my comments: What do we do for web builds?
> 
> Now I'll give the code a run on some platforms.

I'll run some tests to see if the `cursor` CSS property would work fine. Otherwise we may just have to log something like "hardware cursor not supported on web, please resort to hiding it and drawing a sprite at the mouse position". Thoughts?

karl2d.odin
karl-zylinski: perhaps someone writes mad code that destroys and creates them every frame -- that's of course a bad idea, but should it cause an array to grow and grow? Not sure. You could just put in a simple freelist for starters if you want, and not use a handle_map. That the fonts don't do this is an oversight on my end. I'll fix that when I redo the font things

brunodeangelis: added simple freelist

platform_web.odin
```diff
@@ -308,6 +318,43 @@ web_set_window_mode :: proc(window_mode: Window_Mode) {
 	s.window_mode = window_mode
 }
 
+web_create_cursor :: proc(pixels: []Color, width: int, height: int, hotspot: [2]int) -> Cursor_Data {
+	// base64 requires the bytes buffer be PNG encoded, NOT raw pixels.
+	// That's because it needs the metadata from the headers,
+	// otherwise it wouldn't know width height etc.
+	// And so because we're getting raw pixels here, we need to re-encode into PNG.
```
brunodeangelis: @karl-zylinski please take a look at this and the added mini package. 
The PNG encoder I had Claude write, as it's something out of my scope (at least in a timely manner).
I'm not happy with the "dirtiness" of the solution, but pushed it so you could comment on it. Do we either change the API so that we don't need to re-encode, or we push a well-reviewed PNG encoder to the compiler (for `core:image/png`)... or something else.

karl-zylinski: An alternative idea is to only accept a buffer containing a PNG file when you create the cursor. For non-web you use `core:image/png` to read it and extract the data. On web you just pass the PNG data on. Then you make sure to specify in the API docs that it expects a PNG.

brunodeangelis: Check out [the new commit.](https://github.com/karl-zylinski/karl2d/pull/83/changes/087bca650e574378a415b6711e0cc6f87eb97105) It's only for Windows just to see if you'd like that API.
So basically we receive raw bytes that represent an encoded PNG, and each platform will decode as necessary.

karl2d.odin
```diff
@@ -1967,6 +1976,58 @@ get_default_font :: proc() -> Font {
...
+	if cursor.os_handle == nil {
+		return 0
+	}
+	append(&s.cursors, cursor)
+
+	return id
+}
+
+set_cursor :: proc(id: Cursor) {
+	if id == 0 do return
```
karl-zylinski: How do you "unset" the cursor, restoring the default? I'd imagine one would pass `CURSOR_NONE :: Cursor(0)` which would then be used to restore the system's cursor.

karl-zylinski (approved):
Yeah, it looks fine with png-only cursor stuff.

The PR looks OK in other ways too.

When you feel like you are happy with everything, let me know, I'll probably hop onto  the branch and do some minor tweaks here and there.

karl2d.odin
brunodeangelis: Done, now `DEFAULT_CURSOR :: Cursor(0)` can be used to restore the arrow OS cursor temporarily or permanently (depending on if you call `set_cursor` again).

brunodeangelis:
Synced branch with main and solved the merge conflicts

brunodeangelis:
synced to match Mouse locking (#174) and fixed duplicate Wayland bindings

karl-zylinski:
@brunodeangelis I've pushed a lot of code that was written with the help of Claude. I will review it and push some more some day soon. Just a heads up in case you wondered what happened. You don't need to review anything or do anything right now.

In short what changed:
- Some bugs were fixed
- Use Image instead of PNG (with a png encoder like your original idea, I think its better in the end)
- Added load_image_xxx procs etc.

brunodeangelis:
@karl-zylinski thanks for the thorough review, nice to see some tightening up and new features like the OS cursors. Excited to see the final form! Cheers Karl

karl-zylinski:
@brunodeangelis I might merge this tomorrow. I'm just gonna test in on a mac, my mac is a bit sad ATM. I need to install newer LLVM which takes ages (its very old). We'll see how it goes.

What do you think of how it looks?

brunodeangelis:
@karl-zylinski Tested on my M1, works as intended. As for the code, I think it's well managed! There are more features and I like the design. You're probably aware but there are some stuff unrelated to the cursors in the PR, like the mouse locking or the box2d example. Otherwise, LGTM !

karl-zylinski:
Thank you @brunodeangelis . It has now been merged. Yay!!

karl-zylinski:
I forgot to put in the co-author stuff in the description so the squash looks like your commit :D

brunodeangelis:
@karl-zylinski that doesn't seem fair haha! Undo that if you can of course. Thanks for all the help!! 🎉

PR #85, opened 2026-02-21
CoreAudio audio backend

laytan:


platform_bindings/mac/AudioToolbox/AudioToolbox.odin
```diff
@@ -0,0 +1,62 @@
+#+build darwin
+package AudioToolbox
+
+import CA "../coreAudio"
```
karl-zylinski: CoreAudio? Isn't mac case sensitive? I don't understand how this worked when I tested it...

platform_bindings/mac/CoreAudio/CoreAudio.odin
```diff
@@ -0,0 +1,52 @@
...
+}
+FormatFlags :: bit_set[FormatFlag; u32]
+
+StreamPacketDescription :: struct {
+	mStartOffset:            i64,
+	mVariableFramesInPacket: u32,
+	mDataByteSize:           u32,
+}
+
+SMPTETime :: struct {
```
karl-zylinski, line 32: This one seems to be unused.

karl-zylinski:
Also, the checks failed because of some change in the OS make_directory procedure. I updated the master branch with the fix

platform_bindings/mac/AudioToolbox/AudioToolbox.odin
laytan: Case insensitive afaik, but this should be fixed regardless.

platform_bindings/mac/CoreAudio/CoreAudio.odin
laytan: It's used in TimeStamp

PR #93, opened 2026-03-02
add more procs from raylib

brunodeangelis:
only Windows for now until we finalize the API, then I'll do the remaining platforms.
The examples file is a bit random, for now I'm just using it as playground to test all of these.

```
IsKeyPressedRepeat     -> key_went_down_repeat
GetCharPressed         -> get_char_pressed
GetClipboardText       -> get_clipboard_text
IsFileDropped          -> is_file_dropped
LoadDroppedFiles       -> get_dropped_files
UnloadDroppedFiles     -> destroy_dropped_files
CheckCollisionPointRec -> check_rect_point_overlap
IsTextureValid         -> is_texture_valid
```

### DIFF
--- a/platform_windows.odin
+++ b/platform_windows.odin
@@ -16,7 +16,7 @@ PLATFORM_WINDOWS :: Platform_Interface {
 	set_screen_size = windows_set_screen_size,
 	get_window_scale = windows_get_window_scale,
 	set_window_mode = windows_set_window_mode,
-
+	get_clipboard_text = windows_get_clipboard_text,
 	is_gamepad_active = windows_is_gamepad_active,
 	get_gamepad_axis = windows_get_gamepad_axis,
 	set_gamepad_vibration = windows_set_gamepad_vibration,
@@ -25,6 +25,7 @@ PLATFORM_WINDOWS :: Platform_Interface {
 }
 
 import win32 "core:sys/windows"
+import "core:strings"
 import "base:runtime"
 @require import "log"
 
@@ -49,6 +50,7 @@ windows_init :: proc(
 	s.events = make([dynamic]Event, allocator = allocator)
 	s.custom_context = context
 	
+	win32.SetConsoleOutputCP(.UTF8)
 	win32.SetProcessDpiAwarenessContext(win32.DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2)
 	win32.SetProcessDPIAware()
 	CLASS_NAME :: "karl2d"
@@ -93,6 +95,8 @@ windows_init :: proc(
 	windows_set_window_mode(options.window_mode)
 	
 	win32.XInputEnable(true)
+
+	win32.DragAcceptFiles(s.hwnd, true)
 
 	when RENDER_BACKEND_NAME == "d3d11" {
 		s.window_render_glue = {
@@ -340,6 +344,7 @@ Windows_State :: struct {
 	screen_height_before_resize_move: int,
 
 	events: [dynamic]Event,
+	high_surrogate: rune,
 
 	// for when returning from fullscreen to window mode
 	restore_window_pos_x: int,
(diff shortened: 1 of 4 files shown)


platform_windows.odin
```diff
@@ -400,6 +403,20 @@ windows_set_window_mode :: proc(window_mode: Window_Mode) {
...
 
+windows_get_clipboard_text :: proc(allocator := context.allocator) -> string {
+	opened := win32.OpenClipboard(nil)
+	if !opened do return ""
+	defer win32.CloseClipboard()
+
+	handle := win32.GetClipboardData(win32.CF_UNICODETEXT)	
+	if handle == nil do return ""
+
+	ptr := ([^]u8)(win32.GlobalLock((win32.HGLOBAL)(handle)))
```
blob1807: It's possible for `GlobalLock` to fail & return `nil`.

platform_windows.odin
```diff
@@ -400,6 +403,20 @@ windows_set_window_mode :: proc(window_mode: Window_Mode) {
...
+	defer win32.GlobalUnlock((win32.HGLOBAL)(handle))
+
+	return strings.clone_from_bytes(ptr[:win32.GlobalSize(handle)], allocator)
```
blob1807: The text returned is UTF-16 not UTF-8.

platform_windows.odin
```diff
@@ -433,6 +447,37 @@ window_proc :: proc "stdcall" (hwnd: win32.HWND, msg: win32.UINT, wparam: win32.
 			})
 		}
 
+	case win32.WM_SYSCHAR, win32.WM_CHAR:
+		char := rune(wparam)
```
blob1807: The char given is UTF-16 encoded. If it's a surrogate pair, the next `WM_CHAR`/`WM_SYSCHAR` event will have the other half.

platform_windows.odin
```diff
@@ -433,6 +447,37 @@ window_proc :: proc "stdcall" (hwnd: win32.HWND, msg: win32.UINT, wparam: win32.
...
+
+		count := win32.DragQueryFileW(drop, 0xFFFFFFFF, nil, 0)
+		paths := make([]string, count, s.allocator)
+		for i in 0 ..< count {
+			len := win32.DragQueryFileW(drop, i, nil, 0)
+			if len == 0 do continue
+
+			MAX_PATH :: 256
+			buf: [MAX_PATH*2]u16
+			win32.DragQueryFileW(drop, i, raw_data(buf[:]), len + 1)
```
blob1807: You're passing `len + 1` & not `MAX_PATH*2`. The base max path on Windows is 4096, so this could oob.

platform_windows.odin
```diff
@@ -400,6 +403,20 @@ windows_set_window_mode :: proc(window_mode: Window_Mode) {
 	}
 }
 
+windows_get_clipboard_text :: proc(allocator := context.allocator) -> string {
```
blob1807: Before opening the clipboard & getting the data. `IsClipboardFormatAvailable` can be used check if there's data.

platform_windows.odin
```diff
@@ -400,6 +403,20 @@ windows_set_window_mode :: proc(window_mode: Window_Mode) {
...
+	opened := win32.OpenClipboard(nil)
```
blob1807: Whist not need for reading. Should really associate the clipboard with a Window when opening it.

platform_windows.odin
```diff
@@ -433,6 +447,37 @@ window_proc :: proc "stdcall" (hwnd: win32.HWND, msg: win32.UINT, wparam: win32.
...
+			})
+		}
+
+	case win32.WM_DROPFILES:
+		drop := (win32.HDROP)(wparam)
+		point: win32.POINT
+		win32.DragQueryPoint(drop, &point)
+		defer win32.DragFinish(drop)
+
+		count := win32.DragQueryFileW(drop, 0xFFFFFFFF, nil, 0)
```
blob1807, line 498: Returns zero on failure. Probably shouldn't add an event if you can't get the files.

karl-zylinski:
Thanks!

There are a bit too many different features in this PR, so I might never merge it, since I might only want some. It's better if you do each feature as a separate PR, then I might merge some of them. The smaller and more to-the-point PR, the easier it is to review and merge.

brunodeangelis:
Closing in favour of #175 and #176 
This can still be used as reference for when it comes time to implement the Key Went Down Repeat or the Get Char Pressed.

PR #94, opened 2026-03-06
Configurable coordinate system: Y up / Y down

karl-zylinski:
Here are some sketches on making Y up / Y down configurable:

When KARL2D_Y_UP is true, then the origin is bottom left corner of screen and Y goes up. This is nice when using box2D.

Initially I used an setting in `Init_Options`, but I figured that avoiding extra if statements deep in the drawing code is good.


karl2d.odin
```diff
@@ -140,6 +168,10 @@ init :: proc(
 	s.current_shader = s.default_shader
 
 	// FontStash enables us to bake fonts from TTF files on-the-fly.
+	//
+	// Note that FontStash is always set up top-down, regardless of the coordinate system. The text
+	// drawing procedures lay glyphs out top-down and place the finished block themselves. That way
+	// the layout is identical in both coordinate systems.
 	fs.Init(&s.fs, FONT_DEFAULT_ATLAS_SIZE, FONT_DEFAULT_ATLAS_SIZE, .TOPLEFT)
```
karl-zylinski, line 171: Why is this top-left placement used in Y up? Why can't we rely on bottom left and move up? Which way is simpler?

karl2d.odin
```diff
@@ -1414,10 +1498,19 @@ draw_text :: proc(
...
 			} else {
 				invalid_rect_size := Vec2 {font_size*0.5, font_size*0.5}
-				invalid_rect := rect_from_pos_size(position + char_offset + {0, invalid_rect_size.y/2}, invalid_rect_size)
-				
+				offset_from_top := char_offset.y + invalid_rect_size.y/2
+
+				when Y_UP {
+					invalid_y := block_top - offset_from_top - invalid_rect_size.y
+				} else {
+					invalid_y := block_top + offset_from_top
```
karl-zylinski: The word invalid_rect made sense but just Invalid_y alone feels strange

karl2d.odin
```diff
@@ -5237,6 +5396,21 @@ is_typable_rune :: proc(r: rune) -> bool {
 	return r >= 32 && r != 0x7f
 }
 
+// The number of lines `text` occupies. Used to size the text block without having to measure the
+// whole thing: only the height matters for placing the block.
+@(private="file", require_results)
+count_text_lines :: proc "contextless" (text: string) -> int {
```
karl-zylinski, line 5397: Again, not easier to use bottom-left placement? I could be wrong though

PR #112, opened 2026-04-09
[b3] Fix CoreAudio buffer size calculations for stereo samples

moomerman:
Audio examples aren't working on macOS with beta3.

Use `size_of([2]Audio_Sample)` instead of `size_of(Audio_Sample)` in buffer allocation and feed bytes


karl-zylinski:
Thank you! I realize I didn't test macOS before doing the release!

PR #118, opened 2026-04-13
Dual grid tilemap

karl-zylinski:
Similar to https://github.com/karl-zylinski/karl2d/pull/98, but a bit more minimalist.


karl2d.odin
```diff
@@ -3397,6 +3395,11 @@ set_internal_state :: proc(state: ^State) {
 	ab.set_internal_state(s.audio_backend_state)
 }
 
+// Open a URL in the default web browser, if possible.
+open_url :: proc(url: string) {
+	pf.open_url(url)
```
blob1807: Might be worth doing a check to make sure it's actually an URI, or at least that it's not potentially malicious like RL does.

platform_windows.odin
```diff
@@ -308,6 +312,17 @@ windows_get_gamepad_axis :: proc(gamepad: int, axis: Gamepad_Axis) -> f32 {
 	return 0
 }
 
+windows_open_url :: proc(url: string) {
+	cmd_str := fmt.aprintf(
+		"url.dll,FileProtocolHandler %s",
```
blob1807: I don't remember if `url.dll` still does any processing on the URL, but afaik it does just passes it to `ShellExecuteA` anyways.
So using `url.dll` might unnecessary.

karl2d.odin
karl-zylinski: I can check if it starts with `http://` or `https://` -- It's not really possible to check if it goes to a "bad place on the internet"

karl-zylinski:
@blob1807 Thanks for the comments. Trying to make `open_url` safe is a bit iffy. I mean, the person making the program could just as well do a malicious call using `os.process_start` manually.

I could
- Skip having `open_url` to discourage / make it harder to do malicious things.
- Have basic validation that it is a URL

However, I don't think having the procedure is the main problem, the main problem is people downloading random programs. For web build I think it's already pretty safe, since it can't run programs on your computer.

karl2d.odin
blob1807: At least on Windows a scheme is required with the current way you're doing it. So it might just be good to have a scheme check anyways.

blob1807:
Oh, I thinking about it from a different angle. The person making the program not properly sanitizing user input. Providing a basic safety net for them.

IIRC SDL doesn't check anything, and raylib just checks for a `'`. So saying/sign posting it's up to the lib user to properly validate/sanitize, is perfectly fine way of handling it too.

PR #121, opened 2026-04-16
draw_texture_XXX API finalization pass + draw_text rotation

karl-zylinski:
API changes for the `draw_texture` procs so we always have origin and rotation available. Renamed the procs to more descriptive names. I've added deprecated versions of the old procs where possible.

```odin
draw_texture :: proc(
	texture: Texture,
	position: Vec2,
	origin: Vec2 = {},
	rotation: f32 = 0,
	tint := WHITE,
)

draw_texture_section :: proc(
	texture: Texture,
	source: Rect,
	position: Vec2,
	origin: Vec2 = {},
	rotation: f32 = 0,
	tint := WHITE,
)

draw_texture_fit :: proc(
	texture: Texture,
	source: Rect,
	dest: Rect,
	origin: Vec2 = {},
	rotation: f32 = 0,
	tint := WHITE,
)
```

Also added origin + rotation to `draw_text`, inspired by #54

### DIFF
--- a/examples/audio/audio.odin
+++ b/examples/audio/audio.odin
@@ -161,9 +161,10 @@ step :: proc() -> bool {
 		),
 		{20, 20},
 		40,
+		k2.BLACK,
 	)
-	k2.draw_text("Press Space to play a familiar sound.", {20, 200}, 40)
-	k2.draw_text("Press Enter to also play a 1 second 440 hz sine wave.", {20, 240}, 40)
+	k2.draw_text("Press Space to play a familiar sound.", {20, 200}, 40, k2.BLACK)
+	k2.draw_text("Press Enter to also play a 1 second 440 hz sine wave.", {20, 240}, 40, k2.BLACK)
 	k2.present()
 	free_all(context.temp_allocator)
 
--- a/examples/basics/basics.odin
+++ b/examples/basics/basics.odin
@@ -82,10 +82,10 @@ step :: proc() -> bool {
 
 	// k2.color_alpha takes a pre-defined color and replaces the alpha (transparency).
 	k2.draw_rect({4, 95, msg2_width+20, 162}, k2.color_alpha(k2.DARK_GRAY, 192))
-	k2.draw_text("Hellöpe!", {15, 105}, 48, color = k2.LIGHT_RED)
+	k2.draw_text("Hellöpe!", {15, 105}, 48, k2.LIGHT_RED)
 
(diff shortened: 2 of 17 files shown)


karl-zylinski:
After some consideration, I think this new style may not work as well as I thought it would.

```odin
draw_texture :: proc(
	texture: Texture,
	position: Vec2,
	crop: Maybe(Rect) = nil,
	origin: Vec2 = {},
	rotation: f32 = 0,
	tint := WHITE
)
```

I have a game prototype where I used to flip the sprite by negating the source width. Now that goes into this "crop" parameter. That makes not sense. Also, the `atlas, pos, atlas_rect` order is more awkward than `atlas, atlas_rect, pos`, because the `pos` is at the top-left corner of the cropped out part of the texture, which reads more easily using the old format.

```odin
atlas_rect.w *= dir_from_facing(e.facing).x
k2.draw_texture(atlas, pos, atlas_rect, origin, entity_blend_rot(e))
```

I'll think this over some more.

karl-zylinski:
Perhaps something like this is best after all:

```odin
draw_texture :: proc(
    texture: Texture,
    position: Vec2,
    origin: Vec2 = {},
    rotation: f32 = 0,
    tint := WHITE
)

draw_texture_fit :: proc(
    texture: Texture,
    texture_source: Rect
    dest: Rect,
    origin: Vec2 = {},
    rotation: f32 = 0,
    tint := WHITE,
)
```
(or whatever name draw_texture_fit should have) -- I then loose some of the nice handyness that fixed some annoying corner cases I had in raylib, but it is no longer weird: I flip using `texture_source` and `draw_texture` simply cannot flip

karl-zylinski:
Here's the current API design proposal for the draw_texture procs:

```odin
draw_texture :: proc(
	texture: Texture,
	position: Vec2,
	origin: Vec2 = {},
	rotation: f32 = 0,
	tint := WHITE,
)

draw_texture_section :: proc(
	texture: Texture,
	source: Rect,
	position: Vec2,
	origin: Vec2 = {},
	rotation: f32 = 0,
	tint := WHITE,
)

draw_texture_fit :: proc(
	texture: Texture,
	source: Rect,
	dest: Rect,
	origin: Vec2 = {},
	rotation: f32 = 0,
	tint := WHITE,
)
```

It is fairly raylib-like, but with more descriptive procedure names and origin + rotation available as optional parameters for all three.

Issue #3, opened 2026-01-02
DPI scale uniformity


karl-zylinski:
Make sure all platforms behave the same with respects to window DPI scaling.

In Windowed mode the Window should always have the resolution you state. So that the backbuffer gets that resolution. You can resize the window with the DPI scale if you wish. And you can zoom the world / UI with a camera set to the DPI scale if you want it bigger.

## Windows
Works correctly.

## Mac
Not sure

## Linux X11
Not sure

## Linux Wayland
Does not work, it scales automatically by DPI. Wayland doesn't seem to want to not scale automatically. So it might force me to rethink this strategy.
## Web 
Auto scales due to canvas being auto scaled. I think there are workarounds. But try to solve wayland first

## Comments

Since there are some issues above, perhaps I should just always resize window size by DPI when windowed and use full monitor size when in borderless fullscreen. So in windowed mode 1280x720 at 2x DPI scale will give you 2560x1440 backbuffer. But I have a feeling that gamedevs might get annoyed by this later...

PR #125, opened 2026-04-20
Dpi scaling and window terminology

karl-zylinski:
Working on a solution to https://github.com/karl-zylinski/karl2d/issues/3 -- Still just experimenting -- We shall see where it all ends up!


karl-zylinski:
TODO: The dual grid example lags in gl backend for some reasons, investigate why.

karl-zylinski:
Closed because I think this is the wrong direction: It introduces confusing "logical" vs "backbuffer" size that has too much magic. New idea is in #128

PR #128, opened 2026-04-25
DPI scaling: Make all platforms scale window equally

### LINKED ISSUE #3: DPI scale uniformity

karl-zylinski:
This is my current idea for how to handle DPI scaling.

- By default, the screen size you feed into `k2.init` is automatically scaled by the scaling setting in the OS (the "DPI scaling"). This does NO MAGIC: It just makes the window bigger and makes sure that the backbuffer size matches.
-  get_screen_width/get_screen_height will report the scaled-up size. There is no "logical size" vs "render size"
- The programmer can query get_window_scale and use that as a basis for how to scale UI elements etc. For example, you could make a camera that has a zoom equal to the window scale.
- There is `disable_auto_scale_hint` flag in the Init_Options. This will disable the scaling-on-init on windows. On other platforms, such as Linux-Wayland, disabling the scaling is not easily doable because Wayland always does it. Also, on web the scaling is always applied. To undo it you'd have to do a reverse de-scale which will introduce artifacts. So scaling by default is a good idea.

Remaining:
- Linux+X11 still has no scaling support at all
- I don't have a new enough mac to test this

Closes #3


moomerman:
I'm not 100% sure what the expected outcome is here, but on my Mac I have this on my non-HDPI screen:

image

and when I drag to the window to a HDPI screen:

image

To me it looks like the rendering is consistent (ie. the 500x500 green part) but the window itself looks huge on the HDPI screen.  On the non-HDPI screen the window is about 2x the width of the rendered part but on the HDPI screen the window looks about 4x the width of the rendered part.

fearofcode:
On an M1 Mac running Tahoe with a 2.0 scale HDPI display:

Screenshot 2026-04-26 at 9 58 38 AM

As soon as I run any code that uses the 

```
camera := k2.Camera {
	zoom = k2.get_window_scale(),
}
k2.set_camera(camera)
```

idiom this library wants to do going forward with this branch, text shows up pixelated. The JPG texture in `basics.odin` also appears pixelated.

karl-zylinski:
@fearofcode thanks, the pixelation is due to my bad font rendering, it will be fixed later

@moomerman that example is obsolete, so I removed it. I recommend just opening basics example and seeing if it works OK when dragging back and forth

karl-zylinski:
@moomerman @fearofcode To be clear, my thinking is like this:

If you create a 1280x720 window with 1.5x scale, then you'll actually  get a 1920x1080 window and get_screen_width/height will return 1920x1080.

This might seem bad: If you typed 1280, then you may expect to be able to hard-code positions in your game relative to that size.

But that is never the case in real games: If the user enters borderless fullscreen, then any size you initially set is gone anyways. In that case, what is best? Should get_screen_width return some "unscaled" resolution that isn't the borderless fullscreen resolution, or should it just give you the borderless fullscreen resolution? I say that the later is better. This makes it easier to write shaders and such as well, where you are always in "native pixel scale".

So what this PR does is:
- Make windows auto-scale on creation, so 1280x720*1.5 becomes 1920x1080. This is perhaps not ideal, but too many platforms do auto scaling that is hard to get around.
- The true resolution of the backing buffer will always be reported by any get_screen/width/height/size procs
- The programmer wont be confused about scaled sizes, except that they can't assume that the size they sent into `init` can be used to hard-code positions.

Thoughts?

fearofcode:
@karl-zylinski I haven't thought about this issue as much as you have, but it sounds like native pixel scale is the better value to return to callers.

this library is meant for games, not business apps, so we are not obligated to abstract away DPI/true backing buffer resolution the way I believe e.g. web browsers do and we should do what works best for shaders rather than a UI programmer who wants to hard code UI widget positions and have it "work" for some sense of that word regardless of display DPI.

tl;dr lgtm ship it

PR #130, opened 2026-04-25
Hot Reload example

karl-zylinski:
Example that implements hot reload similar to https://github.com/karl-zylinski/odin-raylib-hot-reload-game-template -- One big difference is that I've written a build.odin program instead of relying on scripts.

Not 100% done and tested yet, hence draft.


karl-zylinski:
For now, this has been moved to a separate repository: https://github.com/karl-zylinski/karl2d-hot-reload-template

PR #152, opened 2026-05-03
play_audio_buffer

karl-zylinski:
Play an audio buffer once with playback settings specified. 

Experimental, may need better API design relative to the Sound stuff.

For example, perhaps `Sound` is not needed at all, perhaps we can have only `Audio_Buffer` and `Playing_Audio_Buffer`, where `play_audio_buffer` returns a `Playing_Audio_Buffer`. If someone needs what `Sound` gives, then they could implement that themselves.

### DIFF
--- a/examples/space_cat/space_cat.odin
+++ b/examples/space_cat/space_cat.odin
@@ -515,13 +515,12 @@ update :: proc() {
 			dir = vec2_from_direction[player.dir],
 		})
 
-		// The shoot sound has pitch randomization and spatial panning.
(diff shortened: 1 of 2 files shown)


EriKWDev:
It would be nice to still be able to cancel the sounds manually, so basically still get a handle back but also have it be automatically released when the sound is done. So optionally I can stop it manually and change it's settings, but if I don't do anything it will just one-shot auto play and get removed

karl-zylinski:
Duplicate of https://github.com/karl-zylinski/karl2d/pull/223

PR #167, opened 2026-05-09
Call shutdown() in WASM builds

### LINKED ISSUE #156: `shutdown()` proc never registered with WASM?
For WASM support, the readme says to do this:

```odin
package hello_world

import k2 "karl2d"

main :: proc() {
    init()
    for step() {}
    shutdown()
}

init :: proc() {
    k2.init(1280, 720, "Greetings from Karl2D!")
}

step :: proc() -> bool {
    if !k2.update() {
        return false
    }

    k2.clear(k2.LIGHT_BLUE)
    k2.draw_text("Hellope!", {50, 50}, 100, k2.DARK_BLUE)
    k2.present()

    return true
}

shutdown :: proc() {
    k2.shutdown()
}
```

But it seems that the WASM code generated by `build_web` tool never actually registers the shutdown function. Or, if it does, I can't seem to hit it ever.

hintron:
This fixes #156 and makes it so that `shutdown()` is properly called when a WASM game exits. This is important when trying to use tracking allocators to see if there are any memory leaks.

---

## Rules Checklist

You can always submit a draft pull request. But when you make your Pull Request "ready for review", then please make sure these rules are followed (put an x between each [ ]):
	- Use tabs, not spaces.

### DIFF
--- a/build_web/web_entry_templates/index_template.html
+++ b/build_web/web_entry_templates/index_template.html
@@ -46,6 +46,9 @@
 						prevTimeStamp = currTimeStamp;
 
 						if (!exports.step(dt, odin_ctx)) {
+							if (exports.shutdown) {
+								exports.shutdown(dt, odin_ctx);
+							}
 							if (exports._end) {
 								exports._end();
 							}
--- a/build_web/web_entry_templates/web_entry_template.odin
+++ b/build_web/web_entry_templates/web_entry_template.odin
@@ -20,3 +20,9 @@ step :: proc(dt: f64) -> bool {
 	context = ctx
 	return ex.step()
 }
+
+@export
+shutdown :: proc(dt: f64) {
+	context = ctx
+	ex.shutdown()
+}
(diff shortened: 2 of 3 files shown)


hintron:
Fixes #156

hintron:
Example of it working!:
image

(This is using the magnanimous https://github.com/OnlyXuul/tracker tracking allocator. Looks like there still might be some issues to iron out in WASM, but `tracker` seems to mostly work!)

Code: https://github.com/hintron/traversal/blob/24be41b82019514f2697887aab92b665b137e4f7/src/main.odin#L438-L454

karl-zylinski:
Thanks for the PR.

I understand why you want this, but an important aspect to consider is that this shutdown procedure will never run if a user just closes the tab/window. It's only if you manually exit the game loop within the game. Such "exit game" choices are odd in a web game: What happens after an exit? Does the tab just stay black?

These are questions that made me not put in the shutdown call. We can still have it, but I just want to make sure you understand in which cases shutdown would actually trigger.

hintron:
Right, as far as I can tell it has no practical use for the user of the game, but for the developer, it's useful to be able to set up a tracking allocator that can print out on shutdown.

For my game, I just trigger a shutdown on the escape key. The screen doesn't go black, but rather shows the last frame rendered, and some console errors show up when you do mouse movement. But that's it.

karl-zylinski:
Thanks for your PR

PR #177, opened 2026-05-22
Clarify supported mouse buttons

### LINKED ISSUE #143: Docs say that 255 mouse buttons are supported... But only Left, Right and Middle are
```
// Support for up to 255 mouse buttons. Cast an int to type `Mouse_Button` to use things outside the
// options presented here.
Mouse_Button :: enum {
	Left,
	Right,
	Middle,
	Max = 255,
}
```

Clubhouse1661:
Fixes #143

This updates the `Mouse_Button` docs to match current behavior.

The previous comment implied support for up to 255 mouse buttons, but the current platform backends only report left, right, and middle mouse buttons. This PR clarifies that wording and, `Max` is described as the upper bound of the `Mouse_Button` enum.

This is a docs-only change.

### DIFF
--- a/karl2d.doc.odin
+++ b/karl2d.doc.odin
@@ -1482,8 +1482,8 @@ State :: struct {
 	mix_buffer_offset: int,
 }
 
-// Support for up to 255 mouse buttons. Cast an int to type `Mouse_Button` to use things outside the
-// options presented here.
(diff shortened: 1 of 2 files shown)


karl-zylinski:
Thank you

PR #178, opened 2026-05-22
Add set_window_title API

### LINKED ISSUE #168: allow changing window title during runtime
There's currently no API support for changing the window title, as far as I can tell.

Clubhouse1661:
Fixes #168

This adds `k2.set_window_title(title: string)` so the window title can be changed at runtime.

- added `set_window_title` to the public API
- added `set_window_title` to `Platform_Interface`
- implemented it for Windows, macOS, web, X11, and Wayland
- updated `karl2d.doc.odin`

This is a small API addition. It does not change existing behavior unless the new function is called.

Also, here's a short demo showing the same running window repeatedly changing title between `Alpha` and `Bravo` using `k2.set_window_title(...)`.

https://github.com/user-attachments/assets/fb34c1a5-ad6c-40b9-a047-edd95ee5c5de

### DIFF
--- a/karl2d.doc.odin
+++ b/karl2d.doc.odin
@@ -135,6 +135,9 @@ get_screen_height :: proc() -> int
 // Gets the screen width and height as a 2D vector.
 get_screen_size :: proc() -> Vec2
 
+// Change the window title.
+set_window_title :: proc(title: string)
+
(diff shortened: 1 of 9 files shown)


karl-zylinski:
Thank you!

PR #182, opened 2026-05-24
Add get_window_position API

### LINKED ISSUE #124: Get Current Window Position
We already have get_window_scale but get_window_position doesn't exist yet.

Clubhouse1661:
## Summary

Adds `get_window_position()`.

Fixes #124

## Changes

- Added public `get_window_position() -> Vec2`.
- Added platform interface plumbing.
- Implemented supported platform behavior:
  - Windows returns the visible window position, matching the coordinate space used by `set_window_position`.
  - macOS returns the window frame origin, matching the bottom-left coordinate space used by `set_window_position`.
  - X11 returns the window position translated to root-window coordinates.
- Added unsupported stubs for Wayland and web that log and return `{}`.
- Regenerated `karl2d.doc.odin`.

## Notes

The intended API contract is that `get_window_position()` returns coordinates in the same coordinate system accepted by `set_window_position()` on platforms where window positioning is supported.

Wayland and web do not expose meaningful global window positioning in the same way, so they currently return `{}`.

## Manual Test

On Windows, this smoke test:

```odin
before := k2.get_window_position()
k2.set_window_position(200, 200)
after := k2.get_window_position()

fmt.printf("before: %v\n", before)
fmt.printf("after:  %v\n", after)
```

printed:

```text
before: [40, 32]
after:  [200, 200]
```

### DIFF
--- a/karl2d.doc.odin
+++ b/karl2d.doc.odin
@@ -143,6 +143,11 @@ set_window_title :: proc(title: string)
 // This does nothing for web builds.
 set_window_position :: proc(x: int, y: int)
 
+// Gets the window position in the same coordinate system used by `set_window_position`.
+//
+// This returns {} for web and Wayland builds.
+get_window_position :: proc() -> Vec2
+
 // Fetch the scale of the window. This usually comes from some DPI scaling setting in the OS.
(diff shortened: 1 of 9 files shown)


karl-zylinski:
Thank you

PR #183, opened 2026-05-24
Keep camera projection for render textures

### LINKED ISSUE #139: Setting camera while drawing to render texture breaks texture's scale
It seems calling `k2.set_camera` after `k2.set_render_texture` will always overwrite `proj_matrix`, resulting in texture being scaled relative to window size.

---

Potential patch:
```diff
diff --git a/karl2d.odin b/karl2d.odin
index 1a0648f..db4a439 100644
--- a/karl2d.odin
+++ b/karl2d.odin
@@ -3411,7 +3411,9 @@ set_camera :: proc(camera: Maybe(Camera)) {

        draw_current_batch()
        s.batch_camera = camera
-       s.proj_matrix = make_default_projection(pf.get_screen_width(), pf.get_screen_height())
+       if s.batch_render_target == RENDER_TARGET_NONE {
+               s.proj_matrix = make_default_projection(pf.get_screen_width(), pf.get_screen_height())
+       }

        if c, c_ok := camera.?; c_ok {
                s.view_matrix = camera_view_matrix(c)
```

Clubhouse1661:
Fixes #139.

## Summary

`set_camera` always rebuilt `s.proj_matrix` from the window size. When called after `set_render_texture`, this overwrote the render texture projection, so camera transforms inside render textures were scaled as if they were rendering to the window.

This changes `set_camera` to only reset the projection to the window size when no render texture is active. Render texture projection remains owned by `set_render_texture`.

## Visual confirmation

Before fix:

before_origin_master

After fix:

after_fix

The repro renders the same scene two ways:

- Left: `set_render_texture` then `set_camera`
- Right: `set_camera` then `set_render_texture`

Before the fix, the two panels differ. After the fix, both panels match.

### DIFF
--- a/karl2d.odin
+++ b/karl2d.odin
@@ -3996,7 +3996,10 @@ set_camera :: proc(camera: Maybe(Camera)) {
 
 	draw_current_batch()
 	s.batch_camera = camera
-	s.proj_matrix = make_default_projection(pf.get_screen_width(), pf.get_screen_height())
+
+	if s.batch_render_target == RENDER_TARGET_NONE {
+		s.proj_matrix = make_default_projection(pf.get_screen_width(), pf.get_screen_height())
+	}
 
 	if c, c_ok := camera.?; c_ok {
 		s.view_matrix = camera_view_matrix(c)


karl-zylinski:
Thanks

PR #188, opened 2026-06-01
Several fixes for crashes on Wayland

grekand46:
Someone recently opened an issue about the library crashing on niri (a scroll tiling compositor). This happens because the current wayland implementation assumes that compositors won't send a nonzero size on the first configure, which isn't the case for niri and other tiling compositors that predetermine window sizes, leading to a null pointer being used for the EGL window. The current implementation also doesn't have an optimal configuration sequence for initial window modes and causes a frame of flicker if `options.window_mode` is set to fullscreen. Due to these reasons I thought it would be a better idea to rewrite the configure handler as well as parts of initialization.

Another problem I noticed: the library would also crash when I change back and forth between ttys. This time it's due to an unimplemented `global_removed` handler on the registry. That is fixed now.

I also added some more null checks that hopefully reduces the chances of crashing or at least makes debugging easier. So fat I've tested a few of the examples on niri, labwc, and kwin, and if any other compositor causes more issues I'll try to fix that as well.

---

## Rules Checklist

	- Use tabs, not spaces.

### DIFF
--- a/platform_linux_window_wayland.odin
+++ b/platform_linux_window_wayland.odin
@@ -57,10 +57,20 @@ wl_init :: proc(
 
 	display_registry := wl.display_get_registry(s.display)
 	wl.add_listener(display_registry, &registry_listener, nil)
+
+	// Collects all the globals.
 	wl.display_roundtrip(s.display)
 
 	wl.add_listener(s.seat, &seat_listener, nil)
+
+	// Initializes pointer and keyboard based on seat capabilities.
 	wl.display_roundtrip(s.display)
+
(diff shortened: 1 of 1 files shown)


karl-zylinski:
@grekand46 Thank you for the PR. I pushed some additional fixes, co-created by Claude. See the commit message for details https://github.com/karl-zylinski/karl2d/pull/188/changes/6ac4bf867e129ca0cab8f4c5bc1f67b4c026f5b6

If it looks reasonable to you then I'm happy to merge this.

PR #202, opened 2026-06-27
Prevent segfault

thetarnav:
It's your fix @karl-zylinski from discord
just it's still not on the master branch so I have to keep using my fork


thetarnav:
Looks like https://github.com/karl-zylinski/karl2d/pull/188 already fixes this, gonna switch to using that for now.

PR #215, opened 2026-08-11
Support more WAV variations

### LINKED ISSUE #146: Support more WAV variations
got
```console
Error: Unsupported bits per sample in wav fmt block: 24
```

karl-zylinski:
Fixes #146.

## What was wrong

`load_audio_buffer_from_bytes` rejected a lot of perfectly normal WAV files:

- A fmt chunk had to be exactly 16 bytes, so `WAVE_FORMAT_EXTENSIBLE` (40 bytes) and PCM with a `cbSize` field (18 bytes) both gave "Invalid wav fmt block size". Recording programs write extensible for pretty much anything above 16 bits per sample.
- 24 bit samples gave "Unsupported bits per sample in wav fmt block: 24".
- The RIFF size had to match the file exactly, and some programs don't write it that way.
- Chunk walking stepped four bytes at a time and scanned for chunk ids instead of following the chunk sizes. It happened to work on simple files, but any chunk whose content contained something that looked like an id, or odd sized content, could throw it off.
- A file with no fmt or no data chunk loaded an empty buffer instead of reporting an error.

## What this does

- Walks the chunks properly: read the id and size, use the content of "fmt " and "data", skip everything else, and account for the pad byte after odd sized content. Chunks may come in any order, and metadata chunks (LIST, JUNK, cue, ...) are ignored.
- Accepts any fmt chunk of 16 bytes or more, and follows the sub format GUID of `WAVE_FORMAT_EXTENSIBLE` to find out whether the samples are integers or floats.
- Adds 24 bit integer and 64 bit float samples, so `Raw_Sound_Format` gets `Integer24` and `Float64`.
- Treats the RIFF size as an upper bound: it cuts away trailing junk, but the size of the buffer wins.
- Errors on a missing fmt or data chunk.

The big commented out `Wav_Fmt` block went away, since the fmt chunk is now actually read.

## API change: `Raw_Sound_Format.Float` is `Float32`

`Float` next to the new `Float64` didn't say what it was, so it is `Float32` now. This breaks source compatibility for anyone passing `.Float` to `load_sound_from_bytes_raw` or `load_audio_buffer_from_bytes_raw`. Odin can't deprecate a single enum member, and a duplicate `Float = Float32` member would collide in the switch, so there was no graceful path here. The two examples that used it are updated.

## Testing

Built with `odin run tools/test_examples` (all examples, native and web, with and without `-debug`), and regenerated `karl2d.doc.odin`.

For the loader itself I generated WAV files covering each variation and loaded them through `load_audio_buffer_from_file` with a temporary test program, checking sample rate, channels, sample count and the decoded samples:

- 8, 16, 24 and 32 bit PCM, mono and stereo
- fmt chunk sizes 16, 18 and 40 (extensible, both 24 bit PCM and 32 bit float sub formats)
- 32 and 64 bit float
- extra chunks before, between and after fmt/data, including one with odd sized content
- a RIFF size that doesn't match the file
- rejection cases: no fmt chunk, no data chunk, 12 bits per sample, 5 channels, a 4 byte fmt chunk, and a RIFF file that isn't WAVE

The same sine wave decoded from 16, 24 and 32 bit integer, 32 bit float and 64 bit float all match each other to within 1.5e-05, so the 24 bit path lines up with the ones that already worked. `examples/audio/chord.wav` still loads to the same 56960 samples at 44100 Hz, stereo.

I did not download the `book_drop2.wav` attached to the issue, so it would be good to confirm against that file (and @EriKWDev's files) before merging.

### DIFF
--- a/karl2d.doc.odin
+++ b/karl2d.doc.odin
@@ -1464,7 +1464,7 @@ Raw_Sound_Format :: enum {
 	Integer16,
 	Integer24, // three bytes per sample, little endian
 	Integer32,
-	Float, // 32 bit
+	Float32,
 	Float64,
 }
 
--- a/karl2d.doc.odin
+++ b/karl2d.doc.odin
@@ -524,7 +524,8 @@ set_sound_loop :: proc(sound: Sound, loop: bool)
 // Sounds created using this procedure owns their internal audio buffer: Calling `destroy_sound`
 // will also destroy the audio buffer. 
 //
-// Currently only supports 16 bit WAV files.
+// Supports mono and stereo WAV files with 8, 16, 24 or 32 bit integer samples, or 32 or 64 bit
+// float samples.
 load_sound_from_file :: proc(filename: string) -> Sound
 
 // Load a sound some pre-loaded memory (for example using `#load("sound.wav")`). Returns a `Sound`
@@ -535,9 +536,10 @@ load_sound_from_file :: proc(filename: string) -> Sound
 // Sounds created using this procedure owns their internal audio buffer: Calling `destroy_sound`
 // will also destroy the audio buffer.
 //
-// Currently only supports 16 bit WAV data. Note that the data should be the entire WAV file,
-// including the header. If your data does not include the header, then please use
-// `load_audio_buffer_from_bytes_raw` combined with `create_sound_from_audio_buffer`.
+// Supports mono and stereo WAV data with 8, 16, 24 or 32 bit integer samples, or 32 or 64 bit
+// float samples. Note that the data should be the entire WAV file, including the header. If your
+// data does not include the header, then please use `load_audio_buffer_from_bytes_raw` combined
+// with `create_sound_from_audio_buffer`.
 load_sound_from_bytes :: proc(bytes: []byte) -> Sound
 
 // Load a sound from some raw audio data. You need to specify the data, format and sample rate of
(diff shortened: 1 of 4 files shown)


karl2d.doc.odin
```diff
@@ -1457,10 +1460,12 @@ Audio_Stream_Data :: struct {
 
 // The format used to describe that data passed to `load_sound_from_bytes_raw`.
 Raw_Sound_Format :: enum {
-	Integer8,
+	Integer8, // unsigned, like in 8 bit WAV files. The other integer formats are signed.
 	Integer16,
+	Integer24, // three bytes per sample, little endian
 	Integer32,
-	Float,
+	Float, // 32 bit
```
karl-zylinski: Should probably say Float32 then

karl2d.odin
```diff
@@ -5237,6 +5169,105 @@ is_typable_rune :: proc(r: rune) -> bool {
 	return r >= 32 && r != 0x7f
 }
 
+// The values the `audio_format` field of a wav fmt chunk can have. Extensible means that the real
+// format is in a GUID at the end of the chunk. Recording programs tend to use it for anything with
+// more than 16 bits per sample.
+WAV_FORMAT_PCM :: 1
```
karl-zylinski: If only used in one proc then put in that proc.

karl2d.odin
```diff
@@ -5237,6 +5169,105 @@ is_typable_rune :: proc(r: rune) -> bool {
...
+//	sample_rate:     u32
+//	byte_per_sec:    u32 // sample_rate * byte_per_bloc
+//	byte_per_bloc:   u16 // (num_channels * bits_per_sample) / 8
+//	bits_per_sample: u16
+//
+// Those 16 bytes can be followed by an extension: A u16 with the size of the extension, and for the
+// extensible format also 6 bytes of extra channel information and a 16 byte sub format GUID. The
+// first two bytes of that GUID are the real `audio_format`. We skip the rest of the extension: The
+// channel mask in it only matters for more channels than we support.
+wav_parse_fmt_chunk :: proc(chunk: []u8) -> (
```
karl-zylinski: Unless used twice, inline

PR #227, opened 2026-08-17
Use math.sign to correctly handle wheel delta on wayland

thetarnav:
Previous `delta = value > 0 ? -1 : 1` meant that value=0 would be delta=1, causing mouse scrolling to be jittered—go up and down for no reason.


karl-zylinski:
Thanks

Issue #46, opened 2026-01-18
Re-add depth testing but do it in a better way


karl-zylinski:
I removed depth testing from the lib because it was no longer used for basic 2D rendering.

However, it is good to have because some 2D games like to use it for sorting or some effects.

So I will re-add it to all backends.

We shall see if the default shaders take 2D or 3D position as input. Perhaps it can auto-detect.

As for writing Z value, I need to come up with a plan on how to do that without making any of the pure 2D stuff worse. Either some extra procedure for setting depth or something else. Not sure yet.

karl-zylinski:
Fixed in #229

PR #229, opened 2026-08-17
Re-add depth testing

karl-zylinski:
Re-adds depth testing (#46) behind a single `Init_Options.depth_test` switch, with `set_z` setting the z of subsequent draws — higher z in front, equal z falling back to drawing order so painter's algorithm still works within a layer. z is per-vertex data rather than pipeline state, so `set_z` does not break the batch, no `draw_*` signature changed, and with depth off the vertex stays 20 bytes and rendering is unchanged.

## API changes

- `Init_Options.depth_test: bool` — creates the depth buffers and turns the test on.
- `Init_Options.depth_range_min/depth_range_max: f32` — the z range usable with `set_z`. Both zero gives the default of -1 to 1. Only used when `depth_test` is on.
- `set_z :: proc(z: f32)` — z for subsequent draws. Does not break the batch.
- `get_z :: proc() -> f32`
- `Render_Backend_Interface.get_depth_clip_range: proc() -> (min, max: f32)` — 0..1 for D3D11, -1..1 for GL/WebGL. All four backends implement it.
- The default shaders now declare a `vec3`/`float3` position. With depth off the vertex layout only supplies xy, so the shader sees z = 0.


DallinTG:
@karl_zylinski looks good to me other than a few things

I do not think you need to check if depth is enabled or not
        if s.depth_test {
            (^f32)(&s.vertex_buffer_cpu[base_offset + pos_offset + size_of(Vec2)])^ = s.z
        }
Always setting the value is probably faster and cleaner and shouldn't case issues ”may take testing”

If you do keep the 
if s.depth_test {}
Then i would recommend doing it in the draw procs

 i would recommend having
batch_vertex :: proc(v: Vec2, uv: Vec2, color: Color, batch:^Batch_Draw_Calls) 
Take a Vec3 instead of a vec2 or at the very least having a variant proc like batch_vertex_vec3
Would make it possible to make custom procs that take a vec3 easier for people

I would personally prefer all procs to take a vec3 instead of a vec2 easier and less bug prone but I understand that it is a massive braking change and kinda against the style of a 2d lib so this is a good middle ground. It will lead to bugs where people forget to st it back to zero after setting it and having there stuff get ordered strangely but there is no way of getting around that if you do not want all procs to take a vec3 

There should probably be a way to set the clip space and you shouldn't need to request it from the renderbackend especially if all you are doing is 
gl_get_depth_clip_range :: proc() -> (min: f32, max: f32) {
    return -1, 1
}
Ok looking closer you do have a way to set it but the backend always uses -1,1 and then you use a matrix it smash it into that space i donut know enough to tell you where or not that is the correct choice

karl-zylinski:
@DallinTG The `if s.depth_test {}` is there because the CPU-side vertex is smaller when depth test is off (it's a vec2 when depth test is off). So it is never written in that case.

DallinTG:
O my bad thought it was just always vec3

PR #232, opened 2026-08-18
Seek in sounds

karl-zylinski:
Sounds can now be moved to another spot in their audio. Works the same for sounds played from a clip and from a stream. The audio example got a draggable seek bar.

New API:

```odin
set_sound_time :: proc(sound: Sound, seconds: f32)
get_sound_time :: proc(sound: Sound) -> f32
get_sound_length :: proc(sound: Sound) -> f32
```

Seeking fades the sound out, moves it, then fades it back in, so it does not click. Moving a file stream seeks the file and lets the decoder resynchronize, rather than decoding everything up to the new spot, so it costs a few milliseconds wherever you move to. Stream lengths come from the last ogg page for streams loaded from file, since the decoder cannot report the length of a file it is fed a piece at a time.

Also fixes five things found while testing, all older than this branch:

- On web, everything ran at about twice the speed until the player clicked the page. Browsers play no audio before that, and the backend then reported that nothing was buffered, so the mixer ran once per frame instead of at the rate audio is played. It now counts off what would have been played, using the clock.
- A looping sound clicked when it wrapped round while its volume was ramping. The ramp finished early and then started over in the middle of the mixed chunk.
- A stream was silent for about 0.7 seconds when it started playing. The mixer read the empty buffer before the first `update_audio_stream`, which moved the read position past the write position. That made the buffer look full instead of empty, so it was not refilled until the read position had wrapped all the way around.
- The reported time of a looping stream stopped short of the end and jumped to the start early, since the decoder reads the beginning of the file again while the end of it is still in the buffer waiting to be played.
- Playing a stream that was already playing skipped ahead by however much audio was buffered. It now changes nothing and hands back the sound that is already playing, so the handle you already have stays valid.

### DIFF
--- a/karl2d.odin
+++ b/karl2d.odin
@@ -1979,17 +1979,19 @@ set_sound_position :: proc(sound: Sound, seconds: f32) {
 
 	target_frame := int(clamped_seconds * f32(ab.sample_rate))
 
+	// Don't go past the end of the audio when we know where that is.
+	if sd.total_samples > 0 {
+		target_frame = min(target_frame, sd.total_samples / channels)
+	}
+
 	switch sd.mode {
 	case .From_Bytes:
-		total_frames := int(stbv.stream_length_in_samples(sd.vorbis))
-		clamped_frame := clamp(target_frame, 0, total_frames)
-
-		if stbv.seek(sd.vorbis, u32(clamped_frame)) == 0 {
+		if stbv.seek(sd.vorbis, u32(target_frame)) == 0 {
 			log.error("Cannot set sound position, seeking in the audio stream failed.")
 			return
 		}
 
-		sd.decode_cursor = clamped_frame * channels
+		sd.decode_cursor = target_frame * channels
 		sd.seek_discard = 0
 
 	case .From_File:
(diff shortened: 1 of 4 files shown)


karl2d.odin
```diff
@@ -2777,8 +2935,12 @@ play_audio_stream :: proc(
 		return SOUND_NONE
 	}
 
+	// A stream can only feed one sound, and that sound is already playing, so there is nothing to
+	// start. Hand back the sound that is already going. Unpause it though: This procedure is how
+	// you play a stream, so it should be playing when we are done.
 	if existing := hm.get(&s.sounds, sd.sound); existing != nil {
-		hm.remove(&s.sounds, sd.sound)
+		existing.paused = false
```
karl-zylinski, line 2947: perhaps we should set the settings also

But that requires that the settings are wrapped in Maybe with descriptions of what that means. Otherwise we cant know if someone just passed default explicitly or if its just the default

karl-zylinski:
Position is too easy to mix up with a world position. Find a better name.

Dynamic-Pistol:
Sound timeline?

karl-zylinski:
Renamed to `set_sound_time` / `get_sound_time`. "time" already means seconds elsewhere in the API (`get_time`, `get_frame_time`), and it keeps `position` reserved for the Vec2 kind. `get_sound_length` is unchanged.

The example had a `position` local holding seconds two lines from `get_mouse_position`, which was the same confusion in miniature. That is now `time`.

PR #237, opened 2026-08-25
Balance law for stereo sounds

### LINKED ISSUE #225: Stereo sounds play 3 dB quiet: they use the mono pan placement law
The mixer applies the same constant-power pan curve to every source, regardless of channel count:

```odin
pan_stereo_start := [2]f32 {
    math.cos((pan_start + 1) * math.PI / 4),
    math.sin((pan_start + 1) * math.PI / 4),
}
```

At pan 0 that is `{0.707, 0.707}`. For a mono source that is correct: it is placement, and it keeps `L² + R²` constant so a sound sweeping across the stereo field doesn't dip in the middle.

For a stereo source it is wrong. A stereo file already is a finished stereo mix, so at pan 0 it should come out untouched. Right now it comes out at 0.707 of its authored level.

Measured with a buffer whose samples are all 0.5, mono and stereo side by side:

```
pan      mono L    mono R    stereo L  stereo R
-1.0     0.5000    0.0000    0.5000    0.0000
-0.5     0.4619    0.1913    0.4619    0.1913
 0.0     0.3536    0.3536    0.3536    0.3536
 1.0    -0.0000    0.5000   -0.0000    0.5000
```

Consequences:

- Music is 3 dB quieter than the file. `load_audio_stream_from_xxx` takes the channel count from the ogg, so stereo music goes through this path.
- The doc comment on `set_sound_volume` says "1 is the original volume of the sound". For a stereo sound that is not true.
- miniaudio and SDL_mixer are both transparent at pan 0. miniaudio's own comment on `ma_pan_mode_balance` says it is the default because it is "compatible with other popular audio engines". Anyone coming from raylib, which sits on miniaudio, gets quieter music in Karl2D for no visible reason.

## Suggested fix

Give stereo sources the balance law that the audio buses already use:

```odin
left_gain  = min(1, 1 - pan)
right_gain = min(1, 1 + pan)
```

Mono keeps the constant-power curve. The law follows the signal rather than the object: mono is placement, stereo (a file, or a bus mix) is a finished mix that must pass through untouched when nothing is set.

Note the two laws are already identical at the extremes: both give `{1, 0}` at pan -1. They differ only in the middle, so the change is exactly "stereo stops being attenuated near the center".

Do not fix this by normalizing the constant-power curve so that the center is unity. That would make hard panning boost by 3 dB (`{1.414, 0}`), which can clip. Today nothing in the mixer boosts.

## Cost

A mono file and a dual-mono stereo copy of it would no longer pan identically. That seems worth trading for music playing at the level it was authored at.

`examples/positional_audio` is unaffected: its sources are mono, so they keep the constant-power curve.

karl-zylinski:
Sounds that play a stereo clip now use the balance pan law instead of the constant-power one, so a stereo clip at pan 0 comes out at the level it was authored at rather than 3 dB below it. Mono clips keep the constant-power curve.

Fixes #225.

API changes:

- `set_sound_pan`: doc comment now says what pan does for a mono clip versus a stereo clip.
- `set_audio_bus_pan`: doc comment updated, since the sentence about how sound pan differs from bus pan only applies to mono clips now.

No signature changed, so this is backwards compatible: nothing that compiles today stops compiling. The audible behaviour of stereo clips does change, which is the point of the fix.

Measured with a clip whose samples are all 0.5, before and after:

```
                  before                  after
pan      mono L    mono R    stereo L  stereo R    stereo L  stereo R
-1.0     0.5000    0.0000    0.5000    0.0000      0.5000    0.0000
-0.5     0.4619    0.1913    0.4619    0.1913      0.5000    0.2500
 0.0     0.3536    0.3536    0.3536    0.3536      0.5000    0.5000
 0.5     0.1913    0.4619    0.1913    0.4619      0.2500    0.5000
 1.0    -0.0000    0.5000   -0.0000    0.5000      0.0000    0.5000
```

The mono column is unchanged, and the two laws still agree at the extremes.

What other libraries do, checked against their sources:

- miniaudio: `ma_pan_mode_balance` is the default and is exactly `min(1, 1 - pan)` / `min(1, 1 + pan)`, with a fast path that copies the frames untouched at pan 0.
- SDL2_mixer: `Mix_SetPanning` applies linear per-channel gains and unregisters the effect entirely when both are 255, so the centre is a true no-op.
- Web Audio `StereoPannerNode`: splits on channel count the same way this change does. Mono gets `cos`/`sin` of `(pan + 1) / 2`, which is the same curve Karl2D already used, and stereo passes through untouched at pan 0.

One correction to the issue: raylib is not transparent at the centre. It does not use miniaudio's panner, it has its own law in `MixAudioFrames`, `y = 0.5 * x * (3 - x * x)`, applied to the already-converted stereo frames. At pan 0 that is 0.6875 per channel, so raylib attenuates mono and stereo alike, slightly more than Karl2D did. The rest of the case for the fix stands.

### DIFF
--- a/karl2d.doc.odin
+++ b/karl2d.doc.odin
@@ -562,6 +562,11 @@ sound_is_valid :: proc(sound: Sound) -> bool
 set_sound_volume :: proc(sound: Sound, volume: f32)
 
 // Set the pan of a sound. Range: -1 to 1, where -1 is full left, 0 is center and 1 is full right.
+//
+// What pan does depends on the audio clip. A mono clip is moved between the left and right
+// speakers while keeping its overall loudness the same. A stereo clip is already a finished stereo
+// mix, so instead of moving it, pan turns the opposite side down. Either way, a sound at pan 0
+// plays at the level it was authored at.
 set_sound_pan :: proc(sound: Sound, pan: f32)
 
 // Set the pitch of a sound. Range: 0.01 and up, where 1 is the default. Pitch 2 makes the sound
@@ -713,10 +718,10 @@ set_audio_bus_volume :: proc(bus: Audio_Bus, volume: f32)
 // Set the pan of an audio bus. Range: -1 to 1, where -1 is full left, 0 is center and 1 is full
 // right.
 //
-// This is a balance control: It turns the opposite side down. The pan of a sound works
-// differently: It moves the sound between the left and right speakers while keeping the overall
-// loudness the same. A bus is already a finished stereo mix, and a bus at pan 0 has to leave it
-// exactly as it is.
+// This is a balance control: It turns the opposite side down. A bus is already a finished stereo
+// mix, and a bus at pan 0 has to leave it exactly as it is. A sound that plays a mono clip works
+// differently: It moves between the left and right speakers while keeping the overall loudness the
+// same. See `set_sound_pan`.
 set_audio_bus_pan :: proc(bus: Audio_Bus, pan: f32)
 
 // Set an effect to run on everything that is mixed into the bus. This is how you apply your own
--- a/karl2d.odin
+++ b/karl2d.odin
@@ -2046,6 +2046,11 @@ set_sound_volume :: proc(sound: Sound, volume: f32) {
 }
 
 // Set the pan of a sound. Range: -1 to 1, where -1 is full left, 0 is center and 1 is full right.
+//
+// What pan does depends on the audio clip. A mono clip is moved between the left and right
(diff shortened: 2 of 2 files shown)


karl-zylinski:
Perhaps this is all a bit wrong...

It feels like stereo sounds need two different ways of panning. One that moves the sound from one ear over to the next (-1 means all of left + right channel in left channel). Another one that does what this PR does: Simply lower the volume of the opposite side: -1 is then "left 100% volume right 0% volume".

PR #238, opened 2026-08-25
Mix more than one chunk per update

### LINKED ISSUE #234: Mixer produces one chunk per frame, so audio drops out on long frames
`update_audio_mixer` mixes at most one chunk of `AUDIO_MIX_CHUNK_SIZE` samples per call, and it is called once per frame from `update`. So the most audio the mixer can produce is `AUDIO_MIX_CHUNK_SIZE * frames_per_second` samples per second. With the current 1400 sample chunk and a 44100 sample rate that is 31.5 frames per second. Below that the game cannot make audio as fast as it is played.

The backend only holds `(3 * AUDIO_MIX_CHUNK_SIZE)/2` samples, which is 2100 samples or about 48 milliseconds. Once it runs dry you hear silence, and the mixer can only put 1400 samples back per frame, so it does not catch up quickly either.

Two ways to hit it:

**A frame rate below about 31.** Simulating the mixer loop, the audio produced matches `AUDIO_MIX_CHUNK_SIZE * fps / AUDIO_MIX_SAMPLE_RATE` exactly:

| frame rate | audio produced |
| --- | --- |
| 24 | 0.76x |
| 30 | 0.95x |
| 31.5 | 1.00x |
| 60 | 1.00x |
| 144 | 1.00x |

Above 31.5 it is exact, since the mixer catches up on the next frame whenever it falls behind.

**A single long frame.** This is the one that matters for a game that otherwise runs fine. Level loading, shader compilation or a hitch of any kind:

| one long frame at 60 fps | silence |
| --- | --- |
| 40 ms | none |
| 60 ms | 7.8 ms |
| 80 ms | 27.8 ms |
| 120 ms | 67.8 ms |
| 200 ms | 147.8 ms |

Suggested fix: let the mixer keep producing chunks until the backend has enough, instead of one per call.

```odin
// The nil backend always says zero, so this needs a limit.
MAX_CHUNKS_PER_UPDATE :: 8

for _ in 0..<MAX_CHUNKS_PER_UPDATE {
	if ab.remaining_samples() > (3 * AUDIO_MIX_CHUNK_SIZE)/2 {
		break
	}

	// the body that mixes one chunk
}
```

The nested procedures at the top of `update_audio_mixer` stay where they are, and the part from the buffer wrap check down to `ab.feed(out)` goes in the loop.

This changes how audio is produced on every platform, so it wants its own listening pass. Found while working on #232, which is why it is not in there.

karl-zylinski:
`update_audio_mixer` now keeps mixing chunks until the audio backend has enough of them, instead of stopping after one.

Fixes #234.

No API changes. Backwards compatible.

Measured on Windows/waveout, audio produced per second of wall clock:

| frame rate | before | after |
| --- | --- | --- |
| 16 | 0.48x | 0.77x |
| 28 | 0.83x | 0.94x |
| 32 | 0.88x | 1.01x |
| 62 | 0.98x | 1.01x |

Below about 30 frames per second it still does not reach 1.00x. A frame that long is longer than the (3 * AUDIO_MIX_CHUNK_SIZE)/2 samples the loop tops the backend up to, so the buffer cannot cover it. Raising that target trades latency for it, which the comment in the procedure already asks about.

This changes how audio is produced on every platform. It has not had a listening pass.

### DIFF
--- a/karl2d.odin
+++ b/karl2d.odin
@@ -3242,33 +3242,14 @@ set_audio_bus_effect :: proc(bus: Audio_Bus, effect: Audio_Effect_Proc, user_dat
 //
 // Will only run if the audio backend is running low on audio data.
 update_audio_mixer :: proc() {
-	// If the sample rate of the backend is 44100 samples/second and AUDIO_MIX_CHUNK_SIZE is 1400
-	// samples, then this procedure will only run roughly 44100/1400 = 31 times per second. This
-	// gives a latency of up to (1.5 * (44100/1400)) = 47 milliseconds. Is it too big, or too small?
-	// Perhaps we can use more low latency backends to push it down. Perhaps the backend should
-	// control AUDIO_MIX_CHUNK_SIZE based on how low latency it can give us without stalling?
-	if ab.remaining_samples() > (3 * AUDIO_MIX_CHUNK_SIZE)/2 {
-		return
-	}
-	
-	// We are going to go past the end of the mix_buffer, so just hop to the start instead. It's
-	// 1 megabyte big, so hopping over a few bytes at the end is OK.
-	if (s.mix_buffer_offset + AUDIO_MIX_CHUNK_SIZE) > len(s.mix_buffer) {
-		s.mix_buffer_offset = 0
-	}
-
-	// A slice of the mixed samples we are going to output.
-	out := s.mix_buffer[s.mix_buffer_offset:s.mix_buffer_offset + AUDIO_MIX_CHUNK_SIZE]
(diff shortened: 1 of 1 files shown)


karl-zylinski:
Holding this off... Perhaps a better idea is:
- Add a mixer thread. If mixer thread is enabled, then this is not needed
- If no mixer thread (disabled or on web where there are no threads), then this behavior would happen.

We'd move the current mixing proc to a new proc. The thread and the "multiple chunk processing loop" would both call that proc.

PR #242, opened 2026-08-27
Replace vsync with frame callbacks on wayland

thetarnav:
I had an issue with running karl2d apps where I would get "this app is unresponsive, should I KILL it?" warnings after switching window focus.
Turns out this is due to vsync which caused `egl.SwapBuffers` to block until the window is refocused.

I discovered this when working on clipboard integration, where wayland requires the writer to handle read requests. So when you copy then alt-tab to switch windows to paste, karl2d app would be frozen and the read-request wouldn't be handed until refocused again.

This PR disables vsync on wayland and "reimplements" blocking by polling until a frame callback is called to tell when the surface is ready or until a timeout is exceeded. Which is what glfw does as well: https://github.com/glfw/glfw/blob/92dcf4ce74f2e2554a98fea09be7c705c17daa5a/src/wl_window.c#L2342

It works on my end but I don't know what I'm doing so I would appreciate someone else looking at this.

### DIFF
--- a/platform_bindings/linux/wayland/wayland_core.odin
+++ b/platform_bindings/linux/wayland/wayland_core.odin
@@ -16,6 +16,20 @@ display_flush: proc "c" (display: ^Display) -> c.int
 
 display_dispatch_pending: proc "c" (display: ^Display) -> c.int
 
+display_get_fd: proc "c" (display: ^Display) -> c.int
+
+display_create_queue: proc "c" (display: ^Display) -> ^Event_Queue
+
+event_queue_destroy: proc "c" (queue: ^Event_Queue)
+
+display_dispatch_queue_pending: proc "c" (display: ^Display, queue: ^Event_Queue) -> c.int
+
+display_prepare_read_queue: proc "c" (display: ^Display, queue: ^Event_Queue) -> c.int
+
+display_read_events: proc "c" (display: ^Display) -> c.int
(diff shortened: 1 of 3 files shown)


karl-zylinski:
Thanks for the PR. With some help from Claude I looked into this and fixed some things.

Here's the summary that Claude wrote:
-----
Pushed some changes on top of this.

The main one: the frame callback gets its own event queue now, requested on a proxy wrapper of the surface. Before, waiting for the callback dispatched the default queue, so `xdg_toplevel.configure` landed between the last draw call and `SwapBuffers` and called `egl_window_resize` there. I scripted KWin to resize the window on a timer and counted: 500 of 500 configures arrived mid-frame. Now it is 0 of 499. GLFW and SDL both use a private queue for this.

The rest:

- The timeout is a budget for the whole wait, not per poll. Unrelated events could stretch it indefinitely.
- 20 ms is now 50 ms, as a named constant. 20 ms times out every single frame on anything below 50 Hz.
- `poll` returning -1 retries on EINTR/EAGAIN and bails otherwise. It used to spin forever.
- `display_flush` and the dispatch return values are checked.
- `glFlush` before the wait, so the GPU has something to do.

Tested on KDE Wayland. Visible window sits at 60 fps, minimized drops to 20 fps instead of hanging.

PR #245, opened 2026-08-27
Split init into four procs

karl-zylinski:
Splits `k2.init` into `init_platform`, `open_window`, `init_rendering` and `init_sound`, the same way `k2.update` is already split. `init` and `shutdown` stay as wrappers over them. The point is to let a program do work between the steps, so this also adds monitor queries that work before a window exists.

## API changes

Nothing that existed before changed shape, so no existing game stops compiling. `Init_Options` keeps every field it had: it is now composed from `Window_Options` and `Rendering_Options` with `using`, and Odin lets a struct literal name those subfields directly, so `Init_Options { window_mode = .Windowed_Resizable }` still works. No example needed a change.

New:

- `init_platform`, `open_window`, `init_rendering`, `init_sound`
- `shutdown_sound`, `shutdown_rendering`, `close_window`, `shutdown_platform`
- `get_monitor_count`, `get_monitor_size`, `get_monitor_position`, `get_monitor_scale`
- `Window_Options`, `Rendering_Options`

## Other changes

- The sub-inits are optional. Skipping `init_sound` or `init_rendering` is allowed, and procedures that need a subsystem assert with a message naming the one to call. `update` skips the audio mixer when there is no audio.
- `Platform_Interface` gains `init_process`/`shutdown_process` for setup that has to happen before any window: DPI awareness and the window class on Windows, the `NSApplication` on Mac, the X/Wayland connection and gamepads on Linux. Without this the monitor queries would run in a DPI unaware process and report virtualized sizes.
- Windows enumerates monitors with `EnumDisplayMonitors` into a cached list, primary first, refreshed on `WM_DISPLAYCHANGE`.
- Mac uses `NSScreen.screens`. Web reports the one screen the page runs on.
- X11 reports the whole X screen as a single monitor. Per-monitor enumeration needs XRandR bindings, which do not exist in `platform_bindings/linux/` yet.
- Wayland reports zero monitors. Real support needs a `wl_output` global and listener, which is not written yet.
- New `examples/monitors`, which queries the display before choosing its window size.

### DIFF
--- a/examples/monitors/monitors.odin
+++ b/examples/monitors/monitors.odin
@@ -1,0 +1,96 @@
+// Demonstrates the split of `k2.init` into `init_platform`, `open_window`, `init_rendering` and
+// `init_sound`. Queries the connected monitors right after `init_platform`, before a window
+// exists, and uses that to pick a window size based on the primary monitor.
+package karl2d_monitors_example
+
+import k2 "../.."
+import "core:fmt"
+
+init :: proc() {
+	k2.init_platform()
+
+	monitor_count := k2.get_monitor_count()
+	fmt.printfln("Monitor count: %v", monitor_count)
+
+	for i in 0..<monitor_count {
+		fmt.printfln(
+			"Monitor %v: size %v, position %v, scale %v",
+			i,
+			k2.get_monitor_size(i),
+			k2.get_monitor_position(i),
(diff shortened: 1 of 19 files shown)


karl-zylinski:
Reopened as #250 from a sensibly named branch.

PR #246, opened 2026-08-27
Audio mixer thread

### LINKED ISSUE #234: Mixer produces one chunk per frame, so audio drops out on long frames

karl-zylinski:
Moves the mixer off the frame loop. `_mix_one_chunk` mixes one chunk and is driven either by a background mixer thread (default everywhere but web) or by a loop in `update_audio_mixer` (web, or when the thread is turned off). The audio backend also says how much it wants kept queued, which is where the latency win comes from.

Fixes #234. Supersedes #238.

API changes:
- `Init_Options.disable_audio_mixer_thread` — new field. Zero value keeps the thread on.
- `AUDIO_MIXER_TARGET_SAMPLES` — new constant. Used when the backend has no opinion about how much to keep queued.
- `State` — new `audio_mutex`, `audio_mixer_thread`, `audio_mixer_thread_run`, `audio_mixer_thread_logger` fields.
- `update_audio_mixer` — same signature, now does nothing while the thread runs.

Backwards compatible. Nothing that compiles today stops compiling.

- One `sync.Mutex` guards the four handle maps, the master bus and the mix buffer. The mixer takes it for one chunk, the 28 public audio procs take it for their bodies. `ab.feed` happens outside the lock, because it blocks on both waveout and CoreAudio.
- `sync.Mutex` is not recursive, so five call chains between guarded procs were factored into unguarded `_` helpers.
- `Audio_Backend_Interface` gains `queued_samples`, `target_samples` and `interrupt_feed`. None of it is public API.
- ALSA reads back the buffer it was actually given and asks `snd_pcm_delay` how far behind the device is. The mixer thread keeps that much queued instead of a fixed target. Measured through PipeWire this takes end to end latency from 100 ms to 69 ms.
- `interrupt_feed` releases a `feed` that is waiting on the device, so joining the mixer thread cannot hang on a device that stopped taking samples. Implemented for waveout and CoreAudio, a no-op where `feed` never waits.
- `api_doc_builder` skips declarations whose name starts with an underscore. The new helpers sit next to the code that uses them instead of below `API_END`.
- Merged current master, for the runtime ALSA loading the latency work builds on.

Two things are unresolved. `Audio_Effect_Proc` now runs on the mixer thread with the lock held, so an effect that calls back into Karl2D audio deadlocks, and that restriction is not written into the docs. Hot reload still swaps state out from under a running thread.

Draft because the waveout and CoreAudio paths are only compile checked, and because the listening pass on Windows and macOS is still missing. Linux has been run, measured and listened to. Removing the now redundant ALSA feed thread is #251.


karl-zylinski:
Superseded by #259, which was written from `master` with each decision traced rather than continuing this branch. #259 drops `stop_feeding`, `target_samples` and the shared mixer thread files, all of which only existed to let one Karl2D thread guess what each backend wanted.

The two things worth keeping from here are being carried across: the end of stream fix is tracked in #261, and the fixed stream read buffer is going onto #259 because the resize sits on a path that no longer holds the audio mutex.

Left closed rather than deleted, so the commits stay referenceable.

PR #252, opened 2026-08-28
Window decorations on GNOME/Wayland via libdecor

karl-zylinski:
Draws window decorations with libdecor on compositors that do not decorate windows themselves, which fixes the missing titlebar and borders on GNOME/Wayland. libdecor is loaded at runtime with dlopen, like the Wayland, X11 and ALSA libraries, so it does not become a dependency. Without it the window stays undecorated, which is what happens today. Supersedes #214.

No API changes.

- Server-side decorations are still used when the compositor offers them, so KDE and sway keep the path they use today and never load libdecor. libdecor is only the fallback.
- `KARL2D_LINUX_DECORATIONS=server|libdecor` picks the path by hand, like `KARL2D_LINUX_WINDOWING` does.
- The parts that differ between the two shells sit in `_shell_create`, `_shell_destroy`, `_shell_set_title`, `_shell_set_window_mode` and `_shell_dispatch`. Each one calls a `_shell_libdecor_` or a `_shell_server_` version.
- `libdecor_dispatch` replaces `display_dispatch_pending` on the libdecor path, since it pumps the default queue itself.
- New libdecor bindings in `platform_bindings/linux/libdecor`.
- The `zxdg_toplevel_decoration_v1` bindings were typed against the decoration manager. Fixed.
- The decoration object, the toplevel and the xdg_surface are now destroyed at shutdown. The decoration goes first, or the protocol raises `orphaned`.


karl-zylinski:
Superseded by #260, which requires server-side decorations on Wayland and falls through to X11 where the compositor will not decorate, instead of drawing decorations with libdecor.

Measured on GNOME 49 / Fedora 43, libdecor 0.2.5: with libdecor a 1280x720 window drops from 90 fps to 1 fps when the pointer rests on a decoration edge, with single frames taking several seconds. Stack samples are 88/88 inside libdecor's redraw. Both plugins are affected, for different reasons, and an SDL3 window of the same size on the same machine collapses identically, so this is libdecor rather than Karl2D. The X11 path holds 90 fps at every hover position.

PR #257, opened 2026-08-30
Window icons, and a default Karl2D icon

### LINKED ISSUE #161: Feature: Function allowing to set the runtime window icon.
As in the title. Could either be implemented as initialization option or separate function entirely similar to how raylib handles it.

karl-zylinski:
Adds `set_window_icon`, which sets the icon the OS shows for the game. Implemented on Windows, X11, Wayland, macOS and web. Closes #161.

Karl2D also ships an icon of its own, the K2 mark from the logo, and `init` applies it. A game replaces it by calling `set_window_icon`.

Where Karl2D draws the window frame itself, which on Wayland is what GNOME gets, the icon goes in that titlebar too, in front of the title. GNOME implements no icon protocol, so that titlebar is the only place a game's icon shows there.

## API changes

Three additions. Nothing was changed or removed, so existing code keeps compiling.

- `set_window_icon :: proc(image: Image) -> bool`
- `make_karl2d_icon :: proc(size: int) -> Image` and `make_karl2d_logo :: proc(width: int) -> Image`, which generate the Karl2D icon and logo from pixel art defined in code, at whatever size is asked for

## Other changes

- `init` applies `make_karl2d_icon` as the default icon.
- New example `examples/logo` that generates the logo, uploads it as a texture and draws it.
- Windows: the DIB code in `windows_create_custom_cursor` became `windows_create_hicon`, which now makes both cursors and icons.
- X11: sets `_NET_WM_ICON`. Adds `XChangeProperty` and the atom and mode constants it needs to the bindings.
- Wayland: new `xdg-toplevel-icon-v1` bindings. Warns once about the `.desktop` route when the compositor lacks the protocol. Pads a non-square image, since the protocol rejects those.
- Wayland decorations: the titlebar draws the icon in a box of 16 logical pixels in front of the title, scaled by averaging the source pixels each one covers.
- Wayland decorations: the title is centered on the band its capitals fill. It sat high before, since a font's ascent and descent leave room for accents and descenders that a title rarely has.
- macOS: sets the Dock icon, as windows there have none. Adds `Application_setApplicationIconImage` to `cocoa_extras`.
- Web: sets the favicon from a PNG data URI. `index.html` gets a `karl2d-favicon` link element for it.

## Testing and reviewing
- [x] Platform agnostic code reviewed

Platform specific code reviewed
- [x] Windows
- [x] Mac
- [x] Web
- [x] Linux Wayland
- [x] Linux X11

Tested on:
- [x] Windows
- [ ] Mac
- [x] Web
- [x] Linux Wayland
- [x] Linux Wayland GNOME
- [x] Linux X11


brunodeangelis:
Tested `logo` example on Mac. Seems to work well!

WhatsApp Image 2026-08-31 at 15 42 30 (1)

WhatsApp Image 2026-08-31 at 15 42 30

PR #259, opened 2026-08-30
Run the audio mixer on a thread owned by each backend

### LINKED ISSUE #111: Mixer thread
Investigate making it possible to run audio mixer on a separate thread

Needs some atomics for setting volume etc. Also might need some mutex around the playing sounds array, or perhaps an array of  "new playing sounds" so there is less contention.

karl-zylinski:
The audio backends now have a thread that runs the mixer. This lowers latency and lowers the time spent on the main thread, freeing up time for your gameplay code.

This happens for all platforms except web. On web it works like before. This is because web has no proper threading.

This PR adds mutexes around the audio procedures that affect the state. There is a general audio mutex. There is also a mutex for each audio stream. `update_audio_stream` has been made thread safe so you can run a separate thread that updates your music.

`update_audio_mixer` has been renamed to `update_audio` because it also does some additional house keeping. On web update_audio also runs the mixer. On platforms with an audio thread it just does the housekeeping.

The audio threads are not allowed to allocate, except using temp allocator. But even on temp allocator it is just meant for small fmt print allocations. Care has been put into making sure that the mixing proc (`_mix_audio_into_buffer`) adheres to these rules.

Fixes #111, #234

## Testing and reviewing
- [x] Platform agnostic code reviewed

Platform specific code reviewed:
- [x] Windows
- [x] Mac
- [x] Web
- [x] Linux

Tested on:
- [x] Windows
- [x] Mac
- [x] Web
- [x] Linux

Extra things to test:
- [x] Update audio stream from a thread


audio_backend_alsa.odin
```diff
@@ -96,10 +95,13 @@ alsa_init :: proc(state: rawptr, allocator: runtime.Allocator) {
 	thread.start(s.feed_thread)
 }
 
+// Has the mixer fill the buffer and writes it to the device. `pcm_writei` waits until the device
+// has room, so the device sets the pace and this thread never has to guess at one.
 alsa_thread_proc :: proc(t: ^thread.Thread) {
+	context.allocator, context.logger = _audio_thread_context()
```
karl-zylinski: Why not feed init_context when making thread?

audio_backend_alsa.odin
```diff
@@ -96,10 +95,13 @@ alsa_init :: proc(state: rawptr, allocator: runtime.Allocator) {
...
+
 	for sync.atomic_load(&s.run_thread) {
-		time.sleep(5 * time.Millisecond)
```
karl-zylinski, line 101: Will it spin very fast and eat cpu now?

audio_backend_alsa.odin
karl-zylinski: Deliberate, and I have put the reason in a comment on the thread proc now.

Leaving `init_context` as nil is what makes the runtime hand the thread a temp allocator of its own and destroy it when the thread ends. `_select_context_for_thread` returns `runtime.default_context()` in that case, and `_maybe_destroy_default_temp_allocator` cleans it up. Set `init_context` and that second procedure returns early, so the cleanup becomes ours to do.

All this thread wants is two fields: the game's logger, so what the mixer logs goes where the rest of the game's logging goes, and Karl2D's allocator for anything it allocates. Two lines in the procedure gets that and keeps the automatic cleanup. Going through `init_context` would be four lines at the creation site and would hand the temp allocator back to us.

audio_backend_alsa.odin
karl-zylinski: No, and the sleep going is what makes it exact rather than approximate. Comment added on the thread proc.

`alsa_init` opens the device with a mode of `0`, not `PCM_Open_Mode.NONBLOCK`, so `snd_pcm_writei` waits in the kernel until the device has taken the frames. That call is the pacing. The five millisecond sleep was standing in for it.

It also wakes less often than before. One pass covers `ALSA_BUFFER_SAMPLES`, which is 700 samples, so the thread runs about 63 times a second instead of the 200 the sleep gave it, and each pass mixes a few hundred microseconds worth. The failure paths do not spin either: an underrun goes through `pcm_recover` and back into a blocking write, and a recover that fails clears `run_thread` and leaves the loop.

Worth saying plainly that I cannot run ALSA, so this is from the bindings and the open mode rather than from a machine. It is the first thing to watch when you try it on Linux, and `pcm_delay` is logged every 128 passes if it looks wrong.

audio_backend_alsa.odin
karl-zylinski: You were right and my reason was wrong, so I have done it your way in `abdd8db`.

I claimed leaving `init_context` nil was what gave the thread its own temp allocator. It is not. `global_default_temp_allocator_data` is `@thread_local`, and `_select_context_for_thread` re-points `temp_allocator.data` at it on the thread that is starting, so a thread handed an `init_context` gets its own temp allocator just the same.

The only real difference was cleanup: `_maybe_destroy_default_temp_allocator` returns early when `init_context` is set, so the arena is never freed. One `_destroy_audio_thread_temp_allocator` call as each thread finishes covers that.

`_audio_thread_context` returns a whole `runtime.Context` now, set through `init_context` on waveout and ALSA. CoreAudio still assigns it, because the queue calls that one on a thread of its own that Karl2D never creates, but it is the same procedure and it got shorter too.

PR #266, opened 2026-08-31
fix switching to windowed

brunodeangelis:
My game has a toggle (F11) for Fullscreen/Windowed modes, and it opens in fullscreen from the get go. But once it's up, if I hit F11 to go into Windowed, then the window will be super small and almost invisible (Windows backend). Happens because the backend doesn't record the initial window size for the fullscreen path. This PR fixes that.

---

## Rules Checklist

You can always submit a draft pull request. But when you make your Pull Request "ready for review", then please make sure these rules are followed (put an x between each [ ]):
	- Use tabs, not spaces.

### DIFF
--- a/platform_windows.odin
+++ b/platform_windows.odin
@@ -120,6 +120,11 @@ windows_init :: proc(
 
 	assert(s.hwnd != nil, "Failed creating window")
 
+	// A window that starts in fullscreen has no previous windowed WM_MOVE/WM_SIZE messages from
+	// which to populate the restore rectangle. But the window was initially created at a requested
+	// size, so remember it before expanding to the monitor.
+	windows_capture_windowed_geometry()
+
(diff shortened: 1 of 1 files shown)


karl-zylinski:
Thank you. I did an alternative solution here: https://github.com/karl-zylinski/karl2d/commit/da6fc533d00d2193ae88945393ad6477f1e0ead6

The fix makes sure the size you send into `k2.init` is recorded for restoring to, even when the window mode is initially fullscreen. The commit also fixes a weird thing where the window mode was defaulting to .Windowed until `set_windows_window_mode` was run, so it did weird things in `WM_SIZE` that happened before `set_windows_window_mode` was called.

Note that the thing about the "tiny window" wasn't something I was able to reproduce unless I actually sent in a tiny size into `k2.init`.

Please let me know if all your issue are fixed.

PR #270, opened 2026-09-03
Add main.wasm preload link to html template

thetarnav:
This just makes the browser start downloading the `main.wasm` file when it's parsing the html, instead of when executing the inline js.


karl-zylinski:
Thank you

PR #276, opened 2026-09-09
Add flac decoder for audio clips

mfbulut:
Added a flac decoder to audio clip system

It's a stripped down version of https://github.com/mfbulut/GriPlayer/tree/main/audio/flac

It's tested on https://github.com/sfiera/flac-test-files and its bit-exact againts vendor:miniaudio

llm tools are used for inital implementation, debugging and writing tests, most of the code was written by hand

### DIFF
--- a/flac/flac.odin
+++ b/flac/flac.odin
@@ -1,0 +1,458 @@
+package flac
+
+import "base:intrinsics"
+
+File :: struct {
+	data:      []u8,
+	pos:       u64,
+	bit_buf:   u64,
+	bit_count: u64,
+
+	info:    Info,
+	buf_pos: u64,
+	buf_len: u64,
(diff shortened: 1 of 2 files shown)


karl-zylinski:
Thank for the PR! However, I wonder what the usecase for flac in games are? I have never seen a game use flac. Won't it mainly make music etc bigger without the players of the games noticing any difference? Why would one use it in a game instead of ogg/vorbis?

mfbulut:
I use it instead of wav files it doesn't suffer from generation loss when making edits. I just prefer to save my audio files in this format while in development

As that that it doesn't make sense to ship it to users

Feel free to close the PR if you think its unnecessary, I mainly made to use it myself

